PR #288, opened 2013-07-09
Recusrive call of getValue() -> if the Entity does not hold the identifi...

ThaDafinser:
...er directly

Example is:
Address <-> User (and the Address has the identity)

If you now make an ObjectSelect of User for another entity, currently an error will be raised.

With this fix everything is fine again :-)


gianarb:
I close this issue for inactivity.. It's submitted 2 years ago.. I don't know if it's applicable in this time..
@ThaDafinser do you interested to update this PR? 

Thanks and sorry for delay

PR #301, opened 2013-07-20
Create new objects as required during hydration.

elyobo:
This patch allows the hydrator to create new entities as required during
inflation, if an existing entity cannot be found and there is data where
it expects to be, e.g. instead of

```php
<?php
$new = new Entity;
$new->setFoo('bar');
$data = array(
    'entities' => array($new)
);
$hydrator->hydrate($data, $entity);
?>
```

it is instead possible to do

```php
<?php
$data = array(
    'entities' => array('foo' => 'bar')
);
$hydrator->hydrate($data, $entity);
?>
```

### DIFF
--- a/src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
+++ b/src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
@@ -360,8 +360,9 @@ class DoctrineObject extends AbstractHydrator
                 $value,
                 array_flip($metadata->getIdentifier())
             );
-            $object = $this->find($identifiers, $target) ?: new $target;
-            return $this->hydrate($value, $object);
+            $object   = $this->find($identifiers, $target) ?: new $target;
+            $hydrator = new $this($this->objectManager, $this->byValue);
+            return $hydrator->hydrate($value, $object);
         }
 
         return $this->find($value, $target);
@@ -392,6 +393,7 @@ class DoctrineObject extends AbstractHydrator
         $metadata   = $this->objectManager->getClassMetadata($target);
         $identifier = $metadata->getIdentifier();
         $idkeys     = array_flip($identifier);
+        $hydrator   = null;
 
 
         // If the collection contains identifiers, fetch the objects from database
@@ -400,7 +402,8 @@ class DoctrineObject extends AbstractHydrator
                 // $value is most likely an array of fieldset data
                 $identifiers  = array_intersect_key($value, $idkeys);
                 $item         = $this->find($identifiers, $target) ?: new $target;
-                $collection[] = $this->hydrate($value, $item);
+                $hydrator     = $hydrator ?: new $this($this->objectManager, $this->byValue);
+                $collection[] = $hydrator->hydrate($value, $item);
             } else {
                 $collection[] = $this->find($value, $target);
             }
--- a/src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
+++ b/src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
@@ -389,10 +389,21 @@ class DoctrineObject extends AbstractHydrator
         }
 
         $collection = array();
+        $metadata   = $this->objectManager->getClassMetadata($target);
+        $identifier = $metadata->getIdentifier();
+        $idkeys     = array_flip($identifier);
+
 
         // If the collection contains identifiers, fetch the objects from database
         foreach ($values as $value) {
-            $collection[] = $this->find($value, $target);
+            if (is_array($value) && array_keys($value) != $identifier) {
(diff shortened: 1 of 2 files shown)


src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
```diff
@@ -389,10 +390,23 @@ protected function toMany($object, $collectionName, $target, $values)
...
+
 
         // If the collection contains identifiers, fetch the objects from database
         foreach ($values as $value) {
-            $collection[] = $this->find($value, $target);
+            if (is_array($value) && array_keys($value) != $identifier) {
+                // $value is most likely an array of fieldset data
+                $identifiers  = array_intersect_key($value, $idkeys);
+                $item         = $this->find($identifiers, $target) ?: new $target;
+                $hydrator     = $hydrator ?: new $this($this->objectManager, $this->byValue);
```
MatyCZ: Can you add this part, which automatically set relation to source object? If I use your code, it save NULL value to relation column in DB.

```php
    $targetObject = $hydrator->hydrate($value, $item);
} else {
    $targetObject = $this->find($value, $target);
}

if (null !== $targetObject) {
    $targetField = $this->objectManager->getClassMetadata(get_class($object))->getAssociationMappedByTargetField($collectionName);

    if(!empty($targetField)) {
        $targetMethodName = ucfirst($targetField);
        $targetMethodGet = 'get' . $targetMethodName;
        $targetMethodSet = 'set' . $targetMethodName;

        // Set instance of current object to target object
        if(method_exists($targetObject, $targetMethodGet) && null === $targetObject->$targetMethodGet()) {
            $targetObject->$targetMethodSet($object);
        }
    }

    $collection[] = $targetObject;
}
```

elyobo: @MatyCZ, this is so old now that I'm not sure what this code would do applied to a recent version of the module.  Feel free to make whatever changes you like in your own version but it looks like this change will not be merged upstrem.

src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
```diff
@@ -360,8 +360,9 @@ protected function toOne($target, $value)
                 $value,
                 array_flip($metadata->getIdentifier())
             );
-            $object = $this->find($identifiers, $target) ?: new $target;
-            return $this->hydrate($value, $object);
+            $object   = $this->find($identifiers, $target) ?: new $target;
+            $hydrator = new $this($this->objectManager, $this->byValue);
```
sopta: Hi @elyobo, 
is there any reason why we should create another instance of the DoctrineObject?

elyobo: This is pretty ancient now, but my recollection is that the next line ($hydrator->hydrate()) modified some internal state and that if you didn't instantiate new ones for each hydration weird shit happened.  Or it could have just been that I saw that generally each object got its own hydrator elsewhere, so I duplicated that approach here.  Either way, it seems like this PR is long dead in the water.

sopta: I got it. You're right that internal states are changed every time when I'm calling the hydrate method. On the other hand, the states are always up-to-date, so it's maybe not a big deal.

elyobo: I may have just been erring on the side of caution.  Feel free to go without in your own version and let me know how you get on :)

sopta: I'am currently rewriting our system into Doctrine and I wanna use this Hydrator. I'll stay in touch ;)

gianarb:
Hi! Thanks for this feature, are you sure that the hydrator is the correct author for this flow? :ledger:

src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
elyobo: @sopta See the comments from @bakura10 further up the thread (https://github.com/doctrine/DoctrineModule/pull/301#issuecomment-21991754) as to why I did this; internal states definitely do change.

elyobo:
@gianarb Sorry, I don't understand your question.

gianarb:
Sorry for my english :) 
I don't understand if it's a good feature for this hydrator..
@Ocramius @bakura10 @sopta @danizord Can you help me? :)

gianarb:
Hi! Sorry I close this PR because in this library hydrator are not a primary feature and it creates a lot of bug and problem.. We are deciding how manage it in the future release..

PR #303, opened 2013-07-22
Adds strategies for handling associations during extraction.

elyobo:
The default result of extraction is to leave embedded collections or entities.  In many situations, the desired behaviour is to perform an extraction on the related entities as well.

Two strategies are added to handle this situation (each with byValue and byReference implementations).

* ExtractId returns only the identifier of any associated entities, or an array of identifiers for a collection.
* ExtractRecursive creates new hydrators and extracts any associated entities that it finds.  Infinite recursion (in the case of loops) is prevented by one of two strategies, either by providing a "fallback" strategy (which is used after the first level of recursion) which terminates (e.g. the ExtractId above), or by automatic loop detection, which ensures that the same entity is not processed more than once.


gianarb:
Hi! Thanks for this feature but in this moment we prefer don't merge it.
if you would create a good library to manage integration between DoctrineModule and Zend\Form with this code.. We can add it into composer.json how `suggestion`

PR #308, opened 2013-08-07
Ensure that tryConvertArrayToObject returns an object.

elyobo:
If an ID value is passed which is not found, previously no object would be returned, eventually causing the hydration to return a null instead of an object.

One situation in which this is likely to occur is when an external ID is used as a primary key; in this case, the generator is NONE, but because the id value is provided externally it is passed into the inflation. For new entities, this will not be found, so null is returned and hydration fails.

I'm not completely sure of the implications of this; tests pass, but are there some situations out there where returning an object here instead of null will break things?  Should there be an additional check here to only permit this if the generators for all identifiers are all NONE?

### DIFF
--- a/src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
+++ b/src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
@@ -335,7 +335,7 @@ class DoctrineObject extends AbstractHydrator
             $identifierValues[$identifierName] = $data[$identifierName];
         }
 
-        return $this->find($identifierValues, $metadata->getName());
+        return $this->find($identifierValues, $metadata->getName()) ?: $object;
     }
 
     /**


gianarb:
@elyobo can I close this PR? :)
You problem is fixed into the master branch?

elyobo:
As far as I know this was never fixed in the master. If it is, feel free to
close. Of it's not going to be fixed, feel free to close as will-not-fix as
well :)

gianarb:
Hi @elyobo do you add tests for this feature please?

elyobo:
Yes, the PR includes tests. Unfortunately this PR is a year and three
quarters old, so I'm not sure you can run them!

gianarb:
:) Sorry for delay.. But in this PR you change only one file.. THis PR doesn't include tests o.O

elyobo:
Oh, sorry, I just replied to the email and mistook it for another doctrine
PR! I'll have a look tomorrow, it's late here.

gianarb:
Thanks1

elyobo:
Hi @gianarb sorry for the delay, I can see that this has been fixed upstream now, so feel free to close.  

The fix to hydrateByValue was already made before this pull request and the fix to hydrateByReference was made in e85b86e6 a couple of months after this pull request.  Both fixes leave the tryConvertArrayToObject broken and work around it in the hydrate* methods instead, but it works ;)

gianarb:
Thanks I close this issue! :+1:

PR #382, opened 2014-02-15
Added ability for Hydrator to properly cast custom DBAL types.

jeffturcotte:
This enhancement adds support for proper type conversion during hydration when using custom DBAL types. A custom type would implement `DoctrineModule\Stdlib\Hydrator\TypeConversionInterface` for the conversion to occur.


gianarb:
I close this PR.. Thanks for this feature but in this moment we are not interested to integrate another dependency in this module..

PR #399, opened 2014-03-19
lifetime for zend cache storage

jmleroux:
After a failure with ZF2, this is a proposal for accepting the `$lifeTime` parameter with `ZendStorageCache`.

You can read [the PRs 6001 and 5386](https://github.com/zendframework/zf2/pull/6001) for the full story :wink:

### DIFF
--- a/src/DoctrineModule/Cache/ZendStorageCache.php
+++ b/src/DoctrineModule/Cache/ZendStorageCache.php
@@ -72,8 +72,19 @@ class ZendStorageCache extends CacheProvider
      */
     protected function doSave($id, $data, $lifeTime = false)
     {
-        // @todo check if lifetime can be set
-        return $this->storage->setItem($id, $data);
+        if ($lifeTime) {
+            // TTL is per-storage in Zend\Cache\Storage\StorageInterface
+            $options = $this->storage->getOptions();
+            $oldTtl = $options->getTtl();
+            $options->setTtl($lifeTime);
+            $this->storage->setOptions($options);
+            $result = $this->storage->setItem($key, $value);
+            $options->setTtl($oldTtl);
+            $this->storage->setOptions($options);
+        } else {
+            $result = $this->storage->setItem($id, $data);
+        }
+        return $result;
     }
 
     /**
--- a/src/DoctrineModule/Cache/ZendStorageCache.php
+++ b/src/DoctrineModule/Cache/ZendStorageCache.php
@@ -72,18 +72,21 @@ class ZendStorageCache extends CacheProvider
      */
     protected function doSave($id, $data, $lifeTime = false)
     {
-        if ($lifeTime) {
+        if ($lifeTime !== false) {
             // TTL is per-storage in Zend\Cache\Storage\StorageInterface
             $options = $this->storage->getOptions();
             $oldTtl = $options->getTtl();
             $options->setTtl($lifeTime);
             $this->storage->setOptions($options);
-            $result = $this->storage->setItem($key, $value);
+        }
+
+        $result = $this->storage->setItem($id, $data);
+
+        if ($lifeTime !== false) {
             $options->setTtl($oldTtl);
             $this->storage->setOptions($options);
-        } else {
-            $result = $this->storage->setItem($id, $data);
         }
+
         return $result;
     }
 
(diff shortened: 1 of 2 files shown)


src/DoctrineModule/Cache/ZendStorageCache.php
```diff
@@ -72,8 +72,22 @@ protected function doContains($id)
...
     protected function doSave($id, $data, $lifeTime = false)
     {
-        // @todo check if lifetime can be set
-        return $this->storage->setItem($id, $data);
+        if ($lifeTime !== false) {
+            // TTL is per-storage in Zend\Cache\Storage\StorageInterface
+            $options = $this->storage->getOptions();
+            $oldTtl = $options->getTtl();
+            $options->setTtl($lifeTime);
+            $this->storage->setOptions($options);
```
marc-mabe: You don't need to set the options back using setOptions()` as it's the same instance

src/DoctrineModule/Cache/ZendStorageCache.php
```diff
@@ -72,8 +72,22 @@ protected function doContains($id)
...
+        }
+
+        $result = $this->storage->setItem($id, $data);
+
+        if ($lifeTime !== false) {
+            $options->setTtl($oldTtl);
+            $this->storage->setOptions($options);
```
marc-mabe: You don't need to set the options back using setOptions()` as it's the same instance

src/DoctrineModule/Cache/ZendStorageCache.php
```diff
@@ -72,8 +72,22 @@ protected function doContains($id)
...
-        return $this->storage->setItem($id, $data);
+        if ($lifeTime !== false) {
+            // TTL is per-storage in Zend\Cache\Storage\StorageInterface
+            $options = $this->storage->getOptions();
+            $oldTtl = $options->getTtl();
+            $options->setTtl($lifeTime);
+            $this->storage->setOptions($options);
+        }
+
+        $result = $this->storage->setItem($id, $data);
```
marc-mabe: You need to wrap this call in a try-cache black and reset the TTL on exception, too.

marc-mabe:
@jmleroux As you already noted ZF2 is leaving it up to the adapter how to handle TTL and therefore it's not part of the interface. Because the doctrine cache defines TTL within the API you have to deal with it and crimp the ZF2 adapters into this design.
-> That means you have to check the given ZF2 adapter for the TTL behavior and only allow compatible adapters:

```php
public function __construct(StorageInterface $storage)
{
    // Check TTL will be handled on write (Per-Item-TTL)
    $capabilities = $storage->getCapabilities();
    if (!$capabilities->getMinTtl() || !$capabilities->getStaticTtl()) {
        throw new IncompatibleStorageAdapterException(/*  .... */);
    }
    $this->storage = $storage;
}
```
* `minTtl` will be 0 on adapters not handling TTL (0 means infinite)
* `staticTtl` is the same as Per-Item-Ttl. This will be false of adapters detecting expire items on read using current time + last modification time + current TTL setting

gianarb:
Possiible reference with #485

gianarb:
ping @jmleroux are you interesting to follow this PR? :+1: 
Sorry for delay

seyfer:
What about my proposed solution in #485?

gianarb:
Hi @sayfer! I can try to find a solution.. Maybe @marc-mabe can help us with your experience with Zend\Cache.. Can we help us with your feedback about #485? :) Thanks

seyfer:
Hi.
For me is most important to respect interface and expected behavior for all cache implementations.  
When I faced with problem setting TTL for` ZendCacheStorage` I just wrote some code and made a pull request which linked above. I'm trying to implement expected behavior and to repeat already implemented caches. I think main problem in current realization is that we haven't access to default TTL from config in cache class, that's why we have so weird solutions where we need first to save default TTL from source, than apply TTL from passed parameters and than set default value back, because next call of cache set method could be without TTL parameter. I think cache classes should have some field for default TTL and setters and getters and this field should be filled on construction (in factory?). It's require to review current code around concrete cache classes and in any case correct TTL setting for `ZendCacheStorage` should be implemented. As I said before I was really confused when faced with not completed code. :)
So, this is what I think about problem and I will be satisfied with any solution that will work and will work as expected.

gianarb:
This PR is too old right know, I just close it.
@jmleroux if you are interested just open a new one or ping me to reopen this

Thanks

PR #408, opened 2014-04-29
Incude Zend\Stdlib\Hydrator\Filter\FilterProviderInterface.

gStanev:
Incude Zend\Stdlib\Hydrator\Filter\FilterProviderInterface.This class has an usage in the file below (as FilterProviderInterface), therefore has to be imported


gianarb:
I close this issue
Thanks!

PR #412, opened 2014-05-09
How to set a cache adapter

Gamempire:
How to set a cache adapter, example with pac and memcached


bakura10:
I have made the PR against DoctrineORMModule (https://github.com/doctrine/DoctrineORMModule/pull/373).

PR #417, opened 2014-06-05
Extract discriminator column

guilherme-santos:
When you extract and object that use inheritance you lost the information about which object is, this PR return the discrimination column with the type of object


gianarb:
Hi @guilherme-santos thanks for your help and sorry for delay..
In this moment we prefer don't merge this PR.. We keep this limitation and we can add it into the docs

PR #446, opened 2014-11-01
Added functionality to support custom valueOption attributes

matwright:
A useful feature with ZF2 form elements that have valueOptions is being able to set custom attributes on each valueOption.
In the following example the custom attribute 'data-key' is set on each valueOption.
```
$select=new \Zend\Form\Element\Select('test');
$select->setValueOptions(
    [
        ['attributes'=>['data-key'=>'value'],'value'=>'myValue','label'=>'myLabel'],
        ['attributes'=>['data-key'=>'value2'],'value'=>'myValue2','label'=>'myLabel2']
    ]

    );

echo $this->formselect($select);
```
the element view helper returns the following HTML:
```
<select name="test">
<option value="myValue" data-key="value">myLabel</option>
<option value="myValue2" data-key="value2">myLabel2</option>
</select>
```
Being able to add these attributes to valueoptions is extremely useful for passing additional values for front-end processing, like adding a boolean value for an objects status, category or timestamp, etc.

DoctrineModule currently only supports label and value attributes for its ObjectSelect feature. This PR includes functionality to define custom attributes using a option_attributes option.

The attributes would be defined using a new  'option_attributes' key :
```
$os->setOptions([
            'object_manager' => $odm,
            'target_class' => '\Entity\SomeEntity',
            'property' => 'name',
            'option_attributes' => [
            [ 'data-type'=>'getType']
            ],
            'find_method' => [
                'name' => 'findAll',
                'params' => []
           ]
        ]);
```
Whereby the key option_attributes accepts an array of key/values, the keys represent valid HTML attributes(these are already filtered in ZF2 Select) they typically include  attributes such as data-, aria-, onEvent, etc. The value (getType in the above code example) is a method which will be called on the object and used for the option attribute value.

```
data-key="value"
```

### DIFF
--- a/src/DoctrineModule/Form/Element/Proxy.php
+++ b/src/DoctrineModule/Form/Element/Proxy.php
@@ -56,6 +56,11 @@ class Proxy implements ObjectManagerAwareInterface
     protected $property;
 
     /**
+     * @var array
+     */
+    protected $attributes = array();
+
+    /**
      * @var callable $labelGenerator A callable used to create a label based on an item in the collection an Entity
      */
     protected $labelGenerator;
@@ -153,6 +158,22 @@ class Proxy implements ObjectManagerAwareInterface
     public function getEmptyItemLabel()
     {
         return $this->emptyItemLabel;
+    }
+
+    /**
+     * @return array $attributes
+     */
+    public function getAttributes()
+    {
+        return $this->attributes;
+    }
+
+    /**
+     * @param array $attributes
+     */
+    public function setAttributes(array $attributes)
+    {
+        $this->attributes = $attributes;
     }
 
     /**
@@ -455,6 +476,7 @@ class Proxy implements ObjectManagerAwareInterface
         $identifier = $metadata->getIdentifierFieldNames();
         $objects    = $this->getObjects();
         $options    = array();
+        $attributes=array();
 
         if ($this->displayEmptyItem) {
             $options[''] = $this->getEmptyItemLabel();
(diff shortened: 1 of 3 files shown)


matwright:
@Ocramius would it be possible to get some feedback on this PR please ?

src/DoctrineModule/Form/Element/Proxy.php
```diff
@@ -502,7 +528,19 @@ protected function loadValueOptions()
...
+                if (!is_callable(array($object, $methodName))) {
+
+                        throw new RuntimeException(
+                            sprintf('Method "%s::%s" is not callable', $this->targetClass, $methodName)
+                        );
+                }
+                    $optionAttributes[key($optionAttribute)] =(string) $object->{$methodName}();
+            }
+
+                $options[] = array('label' => $label, 'value' => $value, 'attributes'=>$optionAttributes);
```
jaapio: needs less indenting

matwright: @jaapio I've cleaned up with cs-fixer

matwright:
Can this be merged ?

gianarb:
Thanks @matwright 
Merged a12fcd0be0c2933c387985c72af6942581621023

Can you create a new paragraph here with this feature?? :+1: 
https://github.com/doctrine/DoctrineModule/blob/master/docs/form-element.md

matwright:
Thanks @gianarb sure will add some instructions to the docs over the next few days.

PR #459, opened 2014-12-09
Fix for hydration with 'empty' identifiers

### LINKED ISSUE #456: Hydrators with 'empty' identifiers will not be loaded
I have a identifier, in which it is entirely valid (and possible) for an identifier to have the value '0'. When it is not '0', the DoctrineObject hydrator works as it should, and it will fetch the corresponding object from storage.

However, when the identifier is '0', it will not fetch the existing object from storage. I have traced this back to the following line: https://github.com/doctrine/DoctrineModule/blob/master/src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php#L357

That line checks if the identifier is empty, and PHP evaluates `empty('0')` as `true`. Hence, it will not continue to try to fetch the object from storage. I think that changing the check to something like `strlen($data[$identifierName]) == 0` would be a better option.

kokx:
Should fix #456

### DIFF
--- a/src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
+++ b/src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
@@ -354,7 +354,7 @@ class DoctrineObject extends AbstractHydrator
         }
 
         foreach ($identifierNames as $identifierName) {
-            if (!isset($data[$identifierName]) || empty($data[$identifierName])) {
+            if (!isset($data[$identifierName])) {
                 return $object;
             }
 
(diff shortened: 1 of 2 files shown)


src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
```diff
@@ -354,7 +354,7 @@ protected function tryConvertArrayToObject($data, $object)
         }
 
         foreach ($identifierNames as $identifierName) {
-            if (!isset($data[$identifierName]) || empty($data[$identifierName])) {
+            if (!isset($data[$identifierName])) {
```
bakura10: Sounds sane. @Ocramius ?

gianarb:
Merged 96335d7

Thanks!

PR #468, opened 2015-01-26
Fix#467 bypass value to validate message

fabiopaiva:

PR #479, opened 2015-02-28
Fix #467 Uniqueobject passbyvalue

### LINKED ISSUE #467: UniqueObject Validator printing 'array' in response message
The template message is populated with value, but the value was converted in a array.
eg:
There is already another object matching 'array ( 'name' => 'duplicateValue', )'

Here the value is converted from string to array
https://github.com/doctrine/DoctrineModule/blob/1fb289adf1d82f26eed2387ff74e62bac14df255/src/DoctrineModule/Validator/UniqueObject.php#L107

jaapio:
This pr cleans the pr #468 made by @fabaopaiva for #467


gianarb:
Thanks!! :+1:

Issue #484, opened 2015-03-19
Doctrine ZendStorageCache not sets custom cache ttl in save() method and uses ttl from cache settings


seyfer:
Hello

I try to use cache with Doctrine. And I'm using Redis as cache storage and have this initialization in factory

```
class CacheRedisFactory implements FactoryInterface
{
    public function createService(ServiceLocatorInterface $serviceLocator)
    {
        $cache = new ZendStorageCache($serviceLocator->get('redis'));
        return $cache;
    }
}
```

And when in test controller I try set custom ttl for key like in this code

```
        $siteRepo = $this->getEntityManager()->getRepository(\AppModel\Entity\Site::class);

/* @var $cache \DoctrineModule\Cache\ZendStorageCache */
        $cache        = $this->getEntityManager()->getConfiguration()->getResultCacheImpl();
        $cacheItemKey = 'sites';

$items = $siteRepo->findSites(); // retrieve item from repository
            $cache->save($cacheItemKey, $items, 10); // save item to cache
```

It saves in Redis not for 10 seconds, but for 7200 from my Zend cache settings

```
adapter' => [
                'name'    => 'redis',
                'options' => [
                    //2 hours
                    'ttl'        => 7200,
...
```

Why I can't set custom ttl for custom result set ?

seyfer:
I'm also tried this in repository method. You can see it commented.

```
public function findSites()
    {
        $querybuilder = $this->getEntityManager()
                             ->getRepository($this->getEntityName())
                             ->createQueryBuilder('e');

        $query = $querybuilder->select('e')
                              ->groupBy('e.name')
                              ->orderBy('e.id', 'ASC')
                              ->getQuery();


        //1
        //        $query->useQueryCache(TRUE);
        //        $query->setResultCacheLifetime(3600);
        //        $query->setResultCacheId('sites');

        //2
        //        $query->useResultCache(true, 3600, 'sites');

        return $query->getResult();
    }
```

And it's not work too. It's not save cache.

seyfer:
I found this code in `\DoctrineModule\Cache\ZendStorageCache `

```
/**
     * {@inheritDoc}
     */
    protected function doSave($id, $data, $lifeTime = false)
    {
        // @todo check if lifetime can be set
        return $this->storage->setItem($id, $data);
    }
```

This is the reason why it's not sets. It's a shame to release 0.8 version of module with not completed code. :(

It can be fixed in this way

```
[...]

PR #485, opened 2015-03-19
Fix ZendStorageCache::doSave with lifetime

seyfer:
https://github.com/doctrine/DoctrineModule/issues/484

### DIFF
--- a/src/DoctrineModule/Cache/ZendStorageCache.php
+++ b/src/DoctrineModule/Cache/ZendStorageCache.php
@@ -72,7 +72,10 @@ class ZendStorageCache extends CacheProvider
      */
     protected function doSave($id, $data, $lifeTime = false)
     {
-        // @todo check if lifetime can be set
+        if ($lifeTime) {
+            $this->storage->getOptions()->setTtl($lifeTime);
+        }
+
         return $this->storage->setItem($id, $data);
     }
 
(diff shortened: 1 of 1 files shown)


seyfer:
I added posibility to use ttl from zend cache config for ZendStorageCache.

It's added because Zend cache config have some default ttl time that configured in cache.local.php (or anywhere). If I provide custom ttl it use this value as expiration time, if I provide 0 (default) it sets infinite lifetime. 
And if Zend cache driver used for Doctrine caching I want to use configured in Zend config lifetime. 
So I added a case for -1 value. In this case configured Zend cache value will be used.

For better explanation there is an example:
```
$cache->save($cacheItemKey, $items, 20); // save item to cache
$cache->save($cacheItemKey . "_ne", $items, 0); // save item to cache not expire
$cache->save($cacheItemKey . "_d", $items, -1); // save item to cache with default ttl
```

And if `$cacheItemKey` has value `sites` we will have 3 keys in redis (memcached, etc) storage.
sites - expire in 20 seconds
sites_ne - not expired key
sites_d - use ttl from config

Logic implemented with static properties in `ZendStorageCache`. Because if we change `$this->storage` ttl first time it will be saved in object. So we need save what ttl value was in `$this->storage` before change it. I'm not found possibility to access value from config in another way while in `ZendStorageCache` context.
Than we use this saved default value if lifetime parameter passed with -1 value.

gianarb:
Thanks!

gianarb:
@seyfer can you help me to check this feedback in this PR? [399#issuecomment-73426061](https://github.com/doctrine/DoctrineModule/pull/399#issuecomment-73426061)
What do you think?

gianarb:
Hello @seyfer can we address this? 
Thanks a lot

TomHAnderson:
Closing due to lack of action.

PR #490, opened 2015-04-02
Remove php 5.3 from travis

malukenho:
Fix Continuous integration, removing tests for php 5.3 on travis. On composer we already set to accept only `>=5.4` php version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: php
 
 php:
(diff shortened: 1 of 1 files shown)


gianarb:
Duplicated #469

malukenho:
@gianarb oh right! Closing that

gianarb:
Thanks! :+1:

PR #480, opened 2015-03-02
travis: PHP 7.0 nightly, 5.6 added

TomasVotruba:
PHP 5.3 already about to be removed here https://github.com/doctrine/DoctrineModule/pull/469

PR #491, opened 2015-04-02
Removed obsolete Travis build for 5.3 and added builds for 5.6 and 7.0.

Bilge:
PHP 5.4 is stipulated as the minimum required version in `composer.json` and the builds for 5.3 always fails so I removed the obsolete 5.3 build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,23 +1,28 @@
 language: php
 
 php:
-  - 5.3
   - 5.4
   - 5.5
+  - 5.6
+  - 7.0
   - hhvm
 
 matrix:
   allow_failures:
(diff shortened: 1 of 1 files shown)


gianarb:
Thanks but it's duplicated
#469 
#480

Bilge:
Neither of them are as comprehensive as my changes. One just removes 5.3; doesn't add newer versions. The other adds newer versions but doesn't remove 5.3. Even if you combine both neither of them address the other minor issues that are in my PR including `composer update` being called instead of `composer install` and poorly written shell commands.

gianarb:
Merged thanks!

PR #498, opened 2015-04-22
Update module.config.php

ffgalvao:
Works with php 5.3


gianarb:
Master branch doesn't work with PHP 5.3 because it requires PHP >= 5.4.
I merge this PR anyway for "coherence of syntax"

Thanks

Issue #370, opened 2014-01-21
ObjectSelect with optgroup support


sparrowek:
It would be nice to be able to have an ObjectSelect that groups elements using optgroup html tag.

manuakasam:
@gianarb how would you feel about an option where the optgroup is defined by a second field of the same table? I.e:

```
Items
id | title | group
1 | Foo | foo
2 | FooFoo | foo
3 | Bar | bar
```

And the options then would look similar to this:

```
$this->add([
    'type'    => 'DoctrineModule\Form\Element\ObjectSelect',
    'name'    => 'type',
    'options' => [
        'label'              => 'Items',
        'object_manager'     => $objectManager,
        'target_class'       => 'Entity\Item',
        'display_empty_item' => true,
        'empty_item_label'   => '--- (required) choose an item ---',
        'property' => 'title',
        'optgroup_property' => 'group'
    ]
]);
```

And then stuff being rendered accordingly?

I'll be working this within the current project im working on and could easily do it as a proper doctrineModule feature as well. The thing is that this is a quite "local solution".

It may be cleaner to have a separate table for optgroups and then connect the tables (DB wise).

thoughts?

gianarb:
Hi! In my opinion this is a good idea!

manuakasam:
Alright, I'll give this a shot though then within the coming weeks.

gianarb:
Thanks! :) I wait your PR.. Thanks for this feature

xelax90:
Thanks @manuakasam. I really like this feature. Did you think of a label generator for the optgroups? I would do a PR myself but i'm not that familiar with the doctrine code yet.

manuakasam:
@xelax90 that wouldn't go hand in hand with the current implementation. As you do need an identifier within the database for the current implementation to work. And in this case a label_generator (for the optgroup-title) would not be required. 

The options itself are able to generate their label as usual.

driehle:
Closing for - obviously - lack of interest/relevance.

PR #502, opened 2015-05-01
Feature/optgroup support

manuakasam:
References #370 

@gianarb this would be the first set of feature with which adding is not any danger at all I feel.

However there's still one thing that I wouldn't see as "that awesome".

When your data-structure "optgroup" is empty or null it will create an optgroup that has no title. While this is somewhat OK it would likely be more valid to put those empty optgroup entries within NO optgroup block at all.

A further alternative is to provide yet another option key `optgroup_default` where you could define what happens to empty optgroup elements. I.e. `'optgroup_default' => 'Others'` would name the empty group "Others".

The above functionality is already implemented in my branch. You can check it out here:
https://github.com/manuakasam/DoctrineModule/blob/feature/optgroup-default-container/src/DoctrineModule/Form/Element/Proxy.php#L617-L630

The downside to this is that this is quite a lot of custom code. While fully functional you may consider this too custom and not for a wide enough audience? Just gimme your thouhgt. If it's considered OK, ill write tests for those as well and hit up a follow up PR after this.

### DIFF
--- a/src/DoctrineModule/Form/Element/Proxy.php
+++ b/src/DoctrineModule/Form/Element/Proxy.php
@@ -467,7 +467,8 @@ class Proxy implements ObjectManagerAwareInterface
             return;
         }
 
-        $findMethod = (array)$this->getFindMethod();
+        $findMethod = (array) $this->getFindMethod();
+
         if (!$findMethod) {
             $findMethodName = 'findAll';
             $repository     = $this->objectManager->getRepository($this->targetClass);
--- a/src/DoctrineModule/Form/Element/Proxy.php
+++ b/src/DoctrineModule/Form/Element/Proxy.php
@@ -574,10 +574,30 @@ class Proxy implements ObjectManagerAwareInterface
                         sprintf('Method "%s::%s" is not callable', $this->targetClass, $methodName)
                     );
                 }
+
                 $optionAttributes[key($optionAttribute)] = (string)$object->{$methodName}();
             }
 
-            $options[] = array('label' => $label, 'value' => $value, 'attributes' => $optionAttributes);
+            if (null !== ($optgroupIdentifier = $this->getOptgroupIdentifier())) {
+                $optgroupGetter = 'get' . ucfirst($optgroupIdentifier);
+
+                if (!is_callable(array($object, $optgroupGetter))) {
+                    throw new RuntimeException(
+                        sprintf('Method "%s::%s" is not callable', $this->targetClass, $optgroupGetter)
+                    );
+                }
+
+                $optgroup = $object->{$optgroupGetter}();
+
+                $options[$optgroup]['label']     = $optgroup;
+                $options[$optgroup]['options'][] = array(
+                    'label'      => $label,
+                    'value'      => $value,
+                    'attributes' => $optionAttributes
+                );
+            } else {
+                $options[] = array('label' => $label, 'value' => $value, 'attributes' => $optionAttributes);
+            }
         }
 
         $this->valueOptions = $options;
--- a/src/DoctrineModule/Form/Element/Proxy.php
+++ b/src/DoctrineModule/Form/Element/Proxy.php
@@ -86,12 +86,12 @@ class Proxy implements ObjectManagerAwareInterface
     protected $emptyItemLabel = '';
 
     /**
-     * @var string
+     * @var string|null
      */
     protected $optgroupIdentifier;
 
     /**
-     * @var string
+     * @var string|null
      */
     protected $optgroupDefault;
 
@@ -320,7 +320,7 @@ class Proxy implements ObjectManagerAwareInterface
     }
 
     /**
-     * @return string
+     * @return string|null
      */
     public function getOptgroupIdentifier()
     {
@@ -332,11 +332,11 @@ class Proxy implements ObjectManagerAwareInterface
      */
     public function setOptgroupIdentifier($optgroupIdentifier)
     {
-        $this->optgroupIdentifier = $optgroupIdentifier;
+        $this->optgroupIdentifier = (string) $optgroupIdentifier;
     }
 
     /**
-     * @return string
+     * @return string|null
      */
     public function getOptgroupDefault()
     {
@@ -348,7 +348,7 @@ class Proxy implements ObjectManagerAwareInterface
      */
     public function setOptgroupDefault($optgroupDefault)
     {
-        $this->optgroupDefault = $optgroupDefault;
+        $this->optgroupDefault = (string) $optgroupDefault;
     }
 
     /**
@@ -360,7 +360,7 @@ class Proxy implements ObjectManagerAwareInterface
      */
     public function setIsMethod($method)
     {
-        $this->isMethod = (bool)$method;
+        $this->isMethod = (bool) $method;
 
         return $this;
     }
@@ -582,7 +582,7 @@ class Proxy implements ObjectManagerAwareInterface
                     );
                 }
 
-                $label = (string)$object;
+                $label = (string) $object;
             }
 
             if (count($identifier) > 1) {
@@ -600,7 +600,7 @@ class Proxy implements ObjectManagerAwareInterface
                     );
                 }
 
-                $optionAttributes[key($optionAttribute)] = (string)$object->{$methodName}();
+                $optionAttributes[key($optionAttribute)] = (string) $object->{$methodName}();
             }
 
             // If no optgroup_identifier has been configured, apply default handling and continue
--- a/src/DoctrineModule/Form/Element/Proxy.php
+++ b/src/DoctrineModule/Form/Element/Proxy.php
@@ -19,12 +19,12 @@
 
 namespace DoctrineModule\Form\Element;
 
-use InvalidArgumentException;
-use RuntimeException;
-use ReflectionMethod;
 use Doctrine\Common\Collections\Collection;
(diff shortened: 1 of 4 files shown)


gianarb:
In my opinion is OK! :)

src/DoctrineModule/Form/Element/Proxy.php
```diff
@@ -297,14 +311,31 @@ public function getLabelGenerator()
...
+    {
+        return $this->optgroupIdentifier;
+    }
+
+    /**
+     * @param string $optgroupIdentifier
+     */
+    public function setOptgroupIdentifier($optgroupIdentifier)
+    {
+        $this->optgroupIdentifier = $optgroupIdentifier;
```
danizord: Missing cast to string

src/DoctrineModule/Form/Element/Proxy.php
```diff
@@ -297,14 +311,31 @@ public function getLabelGenerator()
...
      *
-     * @param  boolean         $method
+     * @param  boolean $method
+     *
      * @return Proxy
      */
     public function setIsMethod($method)
     {
-        $this->isMethod = (bool) $method;
+        $this->isMethod = (bool)$method;
```
danizord: Revert this one

src/DoctrineModule/Form/Element/Proxy.php
```diff
@@ -406,7 +442,7 @@ protected function loadObjects()
             return;
         }
 
-        $findMethod = (array) $this->getFindMethod();
+        $findMethod = (array)$this->getFindMethod();
```
danizord: Revert

src/DoctrineModule/Form/Element/Proxy.php
```diff
@@ -519,7 +557,7 @@ protected function loadValueOptions()
                     );
                 }
 
-                $label = (string) $object;
+                $label = (string)$object;
```
danizord: Revert

tests/DoctrineModuleTest/Form/Element/TestAsset/FormObject.php
```diff
@@ -70,7 +75,8 @@ public function __toString()
      */
     public function setId($id)
     {
-        $this->id = (int) $id;
+        $this->id = (int)$id;
```
danizord: Revert

tests/DoctrineModuleTest/Form/Element/TestAsset/FormObject.php
```diff
@@ -81,7 +87,8 @@ public function setId($id)
      */
     public function setEmail($email)
     {
-        $this->email = (string) $email;
+        $this->email = (string)$email;
```
danizord: Revert

tests/DoctrineModuleTest/Form/Element/TestAsset/FormObject.php
```diff
@@ -100,7 +107,8 @@ public function getEmail()
      */
     public function setPassword($password)
     {
-        $this->password = (string) $password;
+        $this->password = (string)$password;
```
danizord: Revert

tests/DoctrineModuleTest/Form/Element/TestAsset/FormObject.php
```diff
@@ -119,7 +127,8 @@ public function getPassword()
      */
     public function setUsername($username)
     {
-        $this->username = (string) $username;
+        $this->username = (string)$username;
```
danizord: Revert

tests/DoctrineModuleTest/Form/Element/TestAsset/FormObject.php
```diff
@@ -138,7 +147,8 @@ public function getUsername()
      */
     public function setFirstname($firstname)
     {
-        $this->firstname = (string) $firstname;
+        $this->firstname = (string)$firstname;
```
danizord: Revert

tests/DoctrineModuleTest/Form/Element/TestAsset/FormObject.php
```diff
@@ -157,7 +167,8 @@ public function getFirstname()
      */
     public function setSurname($surname)
     {
-        $this->surname = (string) $surname;
+        $this->surname = (string)$surname;
```
danizord: Revert

src/DoctrineModule/Form/Element/Proxy.php
```diff
@@ -529,17 +567,37 @@ protected function loadValueOptions()
...
+                $optgroupGetter = 'get' . ucfirst($optgroupIdentifier);
+
+                if (!is_callable(array($object, $optgroupGetter))) {
+                    throw new RuntimeException(
+                        sprintf('Method "%s::%s" is not callable', $this->targetClass, $optgroupGetter)
+                    );
+                }
+
+                $optgroup = $object->{$optgroupGetter}();
+
```
danizord: Should probably check if `$optgroup` is string, or cast it to string.

danizord:
@manuakasam awesome! Thanks for this feature :+1:

src/DoctrineModule/Form/Element/Proxy.php
```diff
@@ -85,6 +85,11 @@ class Proxy implements ObjectManagerAwareInterface
      */
     protected $emptyItemLabel = '';
 
+    /**
+     * @var string
```
danizord: string|null

src/DoctrineModule/Form/Element/Proxy.php
```diff
@@ -297,14 +311,31 @@ public function getLabelGenerator()
     }
 
     /**
+     * @return string
```
danizord: `string|null`

src/DoctrineModule/Form/Element/Proxy.php
manuakasam: Regarding this and related comments: why?
Or more polite: where is the coding standard definition for this. Taking a look at the Proxy class alone sometimes it's done way A sometimes B.

src/DoctrineModule/Form/Element/Proxy.php
manuakasam: I'll certainly revert those, however I'd like to have another opinion on this as the casting isn't defined in either PSR nor the doctrine standards. And a quick irc poll was in favor to this writing style i used :P But as I've mentioned, if there's another say I'll revert.

src/DoctrineModule/Form/Element/Proxy.php
manuakasam: Doesn't go in hand with the rest of the class :\

src/DoctrineModule/Form/Element/Proxy.php
manuakasam: agreed on this place.

manuakasam:
Added the default group handling. Still missing the documentation for that though. Also need more feedback on the proper code-styling of `(string)$foo` vs. `(string) $foo` as well as codeblocks for `@var string` vs. `@var string|null`.

src/DoctrineModule/Form/Element/Proxy.php
danizord: Because you expect `$this->optgroupIdentifier` to be always a string, correct? So there are two ways to ensure that you always set a string to `$this->optgroupIdentifier`:
* Use `is_string()` check and throw exception for invalid arguments (more strict, but more hard to implement and test);
* Cast it to string (easy and painless! It does not solve the problem at all, but it's better than nothing :D).

src/DoctrineModule/Form/Element/Proxy.php
danizord: I'm not sure what's the best way to be honest.

There's a [fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.9/Symfony/CS/Fixer/Symfony/SpacesCastFixer.php) in PHPCSFixer to ensure that there's a single space between cast and variable, but I don't know if there's a fixer to ensure the opposite.

Anyway, I think CS fixes should be done in a separate PR.

src/DoctrineModule/Form/Element/Proxy.php
danizord: If constructor does not force `$this->optgroupIdentifier` to be string, then it can be `null` as well. Not sure if you need to follow the "rest of the class" standard here.

src/DoctrineModule/Form/Element/Proxy.php
manuakasam: Alright agreed then. I'll revert all the casting cs then.

src/DoctrineModule/Form/Element/Proxy.php
manuakasam: Alright, I'll still go with the is_string method. More clear and throws the correct exceptions. Stops people from passing a DateTime object as ident :P

src/DoctrineModule/Form/Element/Proxy.php
gianarb: @manuakasam your last commit (CS_FIX) doesn't fix this problem

gianarb:
Ok! :) 
ping @danizord for the last check :P 
Thanks!

danizord:
:100: :ok_hand:

gianarb:
Thanks @manuakasam for this feature and @danizord for the help! :)

manuakasam:
Thanks for accepting it ;) more to come...

PR #510, opened 2015-05-29
Forgot change version

gianarb:
I'm habitual

ping @Ocramius


gianarb:
@Ocramius in this moment the master branch is WIP to tag 1.0.0 it has a lot of feature and fix we cannot tag all work only with 0.8.2 :) .. 

We can remove Version.php to delete this file into 1.0.0.. IMO

Or we can rebase this commit after 0.8.1 to release 0.8.2 in that point..

gianarb:
Close.. see #517

PR #512, opened 2015-06-04
Bump phpunit version and remove minimum-stability flag

mpalourdio:
The minimum-stability flag seems useless now, please correct me if it's not the case.
phpunit has been bumped to allow ZF 2.5 compliance.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,12 @@ php:
   - 5.4
   - 5.5
   - 5.6
-  - 7.0
+  - 7
   - hhvm
 
 matrix:
   allow_failures:
-    - php: 7.0
+    - php: 7
     - php: hhvm
 
 before_script:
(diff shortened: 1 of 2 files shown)


mpalourdio:
Was minimum-stability set to allow "doctrine/common":  ">=2.4,<2.6-dev" ?
If so, I can cancel the last commit, let me know

gianarb:
@Ocramius have you feedback about this change?
We are working into the master branch for 1.0.0 stable release and this is a good change.. :+1: ~2.4 :)

mpalourdio:
I was wondering why travis could only get zf 2.4.2 (except on hhvm, see travis logs). I just remembered a problem I had yesterday.

Now that the main repository has been splitted, Zend\Ldap requires ext/ldap. As the extension in not avalaible on travis, composer fallbacks to 2.4.2.

Any workaround regarding php extension and travis you are aware of ? We have the same problem with the PR I submitted to DoctrineORMModule unfortunately.

I watched Zend\Ldap travis config, but they didn't have the problem because they get dependencies via ```composer require --dev --no-update```

mpalourdio:
Ok, so I could add a .travis.ini to load ldap.so, seems to work fine, except for php 5.4 that still download ZF 2.4.2

Also, php7 builds fail on the bash check, I suspect the missing quotes around ```hhvm```. Also, isn't weird that the coverage is only on for hhvm ? It doesn't look very judicious to me regarding the build matrix.

I think we should allow coverage for *all* php versions, excepted for hhvm and php7. What do you think ?

mpalourdio:
There it is. I have has a rough time with travis to have all the builds back to green !

mpalourdio:
There it is. I have had a rough time with travis to have all the builds back to green !

mpalourdio:
See here for eventually a workaround before considering a merge https://github.com/zendframework/zf2/issues/7569

mpalourdio:
\o/, this is fixed in https://github.com/zendframework/zf2/pull/7571. Some commits are still relevant anyway. I can reset and/or squash what is worth.

mpalourdio:
I've added a commit that fixes the builds. It is relevant as with phpunit 4.*, hhvm reports a wrong covering percentage, and with php7, xdebug is not loaded, so no coverage at all.

gianarb:
Thanks!

Issue #513, opened 2015-06-09
Feature: Add possibility to provide repository method to options


Wilt:
This should be considered an `enhancement`:
It would be great if the ObjectExists NoObjectExists validator would allow to set a custom repository method to use while validating. This allows user to add additional logic (static) to the fields (dynamic) that are validated.

Right now the `findOneBy` method is by default used to find a match. I could easily imagine an `setRepositoryMethod` method inside the class so that in options the user can provide an options array with the following:

`'repository_method' => 'findOneByMyCustomMethod'`

The `isValid` method would then look like:

```
/**
 * {@inheritDoc}
 */
public function isValid($value)
{
    $value = $this->cleanSearchValue($value);
    if(isset($this->repositoryMethod)){
        $method = $this->repositoryMethod;
    }else{
        $method = 'findOneBy';
    }
    $match = $this->objectRepository->$method($value);

    if (is_object($match)) {
        return true;
    }

    $this->error(self::ERROR_NO_OBJECT_FOUND, $value);

    return false;
}
```

Wilt:
I can make a pull request if there is interest in such a feature

gianarb:
Good IMO.. :) 
can you write a PR with tests and docs improvement?

Thanks a lot!

Wilt:
A pull request can be found here:
https://github.com/doctrine/DoctrineModule/pull/514

TomHAnderson:
See https://github.com/zfcampus/zf-apigility-doctrine/blob/master/src/Server/Validator/ObjectExistsFactory.php#L39 for a working example of your request.

PR #514, opened 2015-06-10
Added custom repository method feature

Wilt:
The code is updated to support a custom repository method. The repository method can be provided in the options array as follows:

```
'repository_method' => 'string'
```

In the constructor the provided value will be validated and if the variable is not a string or a non existing method an `InvalidArgumentException` will be thrown.

If no custom method is provided the class will fallback to the default `findOneBy` method.

Tests are provided

### DIFF
--- a/src/DoctrineModule/Validator/NoObjectExists.php
+++ b/src/DoctrineModule/Validator/NoObjectExists.php
@@ -47,7 +47,8 @@ class NoObjectExists extends ObjectExists
     public function isValid($value)
     {
         $value = $this->cleanSearchValue($value);
-        $match = $this->objectRepository->findOneBy($value);
+        $method = $this->repositoryMethod;
+        $match = $this->objectRepository->$method($value);
 
         if (is_object($match)) {
             $this->error(self::ERROR_OBJECT_FOUND, $value);
--- a/docs/validator.md
+++ b/docs/validator.md
@@ -6,6 +6,7 @@ All three validators accept the following options :
 
 * `object_repository` : an instance of an object repository.
 * `fields` : an array that contains all the fields that are used to check if the entity exists (or does not).
+* `repository_method` (optional) : a string that points to a valid method in the provided object repository. If this value is not provided the default method `findOneBy` will be used.
 
 The `DoctrineModule\Validator\UniqueObject` also needs the following option:
 
@@ -24,7 +25,8 @@ You can directly instantiate a validator the following way:
 ```php
 $validator = new \DoctrineModule\Validator\ObjectExists(array(
     'object_repository' => $objectManager->getRepository('Application\Entity\User'),
-    'fields' => array('email')
+    'fields' => array('email'),
+    'repository_method' => 'findOneByCustomMethod' // Optional
 ));
 
 var_dump($validator->isValid('test@example.com')); // dumps 'true' if an entity matches
@@ -70,7 +72,8 @@ class User extends Form
 
        	$noObjectExistsValidator = new NoObjectExistsValidator(array(
             'object_repository' => $entityManager->getRepository('Application\Entity\User'),
-            'fields'            => 'email'
+            'fields'            => 'email',
+            'repository_method' => 'findOneByCustomMethod' // Optional
        	));
 
        	$emailInput->getValidatorChain()
--- a/src/DoctrineModule/Validator/ObjectExists.php
+++ b/src/DoctrineModule/Validator/ObjectExists.php
@@ -98,15 +98,15 @@ class ObjectExists extends AbstractValidator
         $this->objectRepository = $options['object_repository'];
 
         if (isset($options['repository_method'])) {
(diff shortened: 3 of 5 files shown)


Wilt:
Issue/Feature description with this pull request can be found here:
https://github.com/doctrine/DoctrineModule/issues/513

gianarb:
Thanks can you improve the docs with this feature?
Thanks a lot!

Wilt:
I added docs and I added the custom method feature also to the `NoObjectExists` and `UniqueObject` classes

Wilt:
Will this be merged in the next release...?

gianarb:
Yes I'm waiting feedback but nothing ehehehe

ping @Ocramius @danizord @manuakasam :) Help me with your experience please!

manuakasam:
I think I do know about the use-case, I'm just wondering if it's really one that justifies the added complexity. The way I see this, we create another layer of errors. For example

```php
$val = new ObjectExists([
  //default config
  'repository_function' => 'findOnlyThoseMatchingCriteria'
]);

$val->isValid([
    'is_valid' => 1,
    'is_not_expired' => 1
]);
```

With the above sample code, we now have two places where we have to look for code that potentially modifies the result. Not only do we have conditions within the `isValid()` but we also likely have additional conditions within the `findOnlyThoseMatchingCritera`.

I can see the reasoning behind this feature, though, so I don't have general objections. However I do begin to see reasons for Marcos objections of "userland"-features within the component :)

src/DoctrineModule/Validator/NoObjectExists.php
```diff
@@ -47,7 +47,8 @@ class NoObjectExists extends ObjectExists
     public function isValid($value)
     {
         $value = $this->cleanSearchValue($value);
-        $match = $this->objectRepository->findOneBy($value);
+        $method = $this->repositoryMethod;
+        $match = $this->objectRepository->$method($value);
```
danizord: Should probably check if `method_exists` before calling

danizord: ah, nvm. I see line 101. :D

danizord:
I agree with @manuakasam, I see a lot of people opening issues asking how to have specific validation results for their business logic. But it can be useful for simple cases where you only care if objects are found or not. So, it's ok for me :+1: 

Can you write docs about this limitation and tell people to write their own implementations if they want more specific results?

Wilt:
@danizord Sure I don't mind writing some more documentation, but I don't really understand what you are aiming for in your comment... Can you elaborate?

gianarb:
@Wilt a lot of maintainers (I one of them) have the constant fear that a new line of code between a few time will be unmaintainable :)
 This project has how first target the integration between Doctrine and Zend Framework..
All others stuff require a lot of consideration because there are a lot of projects that use this module and every little unneeded update could be cause of problems :)  
Excuse me for my english and I'm not speaking about this feature in particular this is a general comment :)

danizord:
@Wilt sorry for my bad inglesh :P I was talking about the limitation that @manuakasam was talking.

Something like:
>If you want more specific results like "this option was expired" or "don't have permission to choose this option", you should implement your own validator.

Wilt:
The proposed change does not change default behavior of the validator. I also think the change will not create any issues if people extended the classes for their custom implementation. 

I fully understand the critics on making small changes on a important much used library like this. But right now I notice that I can almost never use the default `ObjectExists` validator class in practice since in my model a validation query is never as simple as `findOneBy`. Most of the time I end up copying/extending the class and adding my custom query. IMO it would be a very useful addition without BC issues.

I would not mind making any additional changes or documentation where requested. I am already using it in my own project but it would anyway be great if it could be merged into the next release.

gianarb:
@Wilt I ready to merge this feature but I have the last request :) 

is a good idea delete this PR and move this feature in a separate validator class, for example CallbackValidator

http://framework.zend.com/manual/current/en/modules/zend.validator.callback.html

Thanks a lot!

Wilt:
I don't understand your proposition. How would I be able to move this functionality into the ZF2 callback validator class? Or do you mean I should create two new classes instead of extending the existing validators?

Wilt:
Today I discovered that configuring a custom method is also possible for [the form element classes](https://github.com/doctrine/DoctrineModule/blob/master/docs/form-element.md#example-3--extended-version). 

Isn't adding this `find_method` to the Validator classes is totally in the same line of design.
Maybe an idea would to make the key names and working exactly the same. Would such a change increase chances of this PR being merged?

I keep trying :)

gianarb:
@Wilt IMO this feedback is very good.. Can you level out this configuration!?

After it I'm ready to merge it and to tag stable release of DoctrineModule!

gianarb:
This PR is too old right know, I just close it 
@Wilt feel free to open a new one or ping me to reopen if you are interested to continue

Thanks!

PR #522, opened 2015-08-21
Fix default cache service name

gianarb:


Ocramius:
This is a BC break, @gianarb :|

gianarb:
Yes.. It is for 1.0.0

gianarb:
This is a BC Break but after a IRC speech we can keep "array"

Issue #506, opened 2015-05-11
Hydrator Field Naming Flaw


sebfek:
I have fields on an entity that are underscored.
The inputs from the inputfilter are underscored too and matches the member variable name of the entity.
I tried by_value with true and false.
The problem is that it not hydrates my_value to my_value - only myvalue to my_value (or myvalue) on the entity.

gianarb:
Hi! Can you paste a lot of code? Thanks

TomHAnderson:
closed; no response.

PR #524, opened 2015-08-24
Better support for snake_case field names.

neobyteUK:
- Affects hydrateByValue and extractByValue.
- foo_var_name becomes setFooVarName instead of setFoo_var_name
- foo_var_name becomes getFooVarName instead of getFoo_var_name
- foo_var_name becomes isFooVarName instead of isFoo_var_name
- Also removes 2 unused uses.


neobyteUK:
This probably fixes #506.

srayner:
Looks like travis-ci is failing because the equals sign on line 184 is not aligned with the equals sign on line 183.

gianarb:
Thanks

PR #541, opened 2015-11-30
Change isNullIdentifier() accessibility to protected

Bilge:
`isNullIdentifier()` is the only private method in this class, and I don't think it's very well justified, particularly since I require it in a derived class that implements support for embedded documents.

### DIFF
--- a/src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
+++ b/src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
@@ -543,7 +543,7 @@ class DoctrineObject extends AbstractHydrator
      *
      * @return bool
      */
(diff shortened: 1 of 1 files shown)


gianarb:
ping @Ocramius  opinion?
Thanks!

Ocramius:
If needed, move to an external collaborator.

Making it protected is not a good idea for bc reasons.

Bilge:
@Ocramius Are you high, bruv? Marking a private method protected never breaks compatibility.

Ocramius:
It doesn't break compatibility, but adds a hidden bc endpoint for when we
want to kill it/replace it.

gianarb:
I close this PR

Thanks

PR #547, opened 2016-01-19
Allow for the use of Zend\Cache\Service\StorageCacheAbstractServiceFactory

BreiteSeite:
## Motivation

ZF2 contains an abstract factory to create caches we want to use. We also want to use the `DoctrineModule\Cache\ZendStorageCache` adapter to use the zend cache storages created from the abstract factory without having to write a new factory just to use the zend cache storages / bridge between zend and doctrine caches.

## Issues

1. The `DoctrineModule\Cache\ZendStorageCache` requires the zend cache storages as constructor argument. The `DoctrineModule\Service\CacheFactory` is hardwired to only pass the zend cache storage as constructor argument if the doctrine-adapter (with the name) `predis` is used.

## Workaround

A workaround is to use `predis` as config-key for the cache and rely that the CacheFactory of DoctrineModule then does the right thing and pass the storage as constructor argument. Obvious drawbacks are:
- only 1 cache storage created from the abstract factory is useable (the one under the key `predis`) and
- it is semantically wrong to - for example - have a memcache instance under the config-key `predis`

### config
```yaml
service_manager:
    abstract_factories:
        - "Zend\\Cache\\Service\\StorageCacheAbstractServiceFactory"
doctrine:
    configuration:
        metadata_cache: "predis"
    cache:
        predis: # notice how we overwrite the pre-configured predis doctrine cache to use a memcache
            class: "DoctrineModule\\Cache\\ZendStorageCache"
            instance: memcache
caches:
    memcache:
        adapter:
            name: "memcache"
            options:
                servers:
                ...
```
## Fix

The `DoctrineModule\Service\CacheFactory` is not hardwired to the name of the cache, but to the class-name which is instantiated. The drawbacks listed above disappear and the config is semantically correct. It also should be more future proof, as relying on name is not robust.

### config
```yaml
service_manager:
    abstract_factories:
        - "Zend\\Cache\\Service\\StorageCacheAbstractServiceFactory"
doctrine:
    configuration:
        metadata_cache: "memcache"
    cache:
        memcache:
            class: "DoctrineModule\\Cache\\ZendStorageCache"
            instance: memcache
caches:
    memcache:
        adapter:
            name: "memcache"
            options:
                servers:
                ...
```

Thank you in advance for considering my pull request. If you have any questions or improvements please let me know and i will update the PR.

### DIFF
--- a/src/DoctrineModule/Service/CacheFactory.php
+++ b/src/DoctrineModule/Service/CacheFactory.php
@@ -62,10 +62,12 @@ class CacheFactory extends AbstractFactory
             case 'Doctrine\Common\Cache\FilesystemCache':
                 $cache = new $class($options->getDirectory());
                 break;
+
             case 'DoctrineModule\Cache\ZendStorageCache':
             case 'Doctrine\Common\Cache\PredisCache':
                 $cache = new $class($instance);
                 break;
+
             default:
                 $cache = new $class;
         }
--- a/src/DoctrineModule/Service/CacheFactory.php
+++ b/src/DoctrineModule/Service/CacheFactory.php
@@ -58,15 +58,14 @@ class CacheFactory extends AbstractFactory
             $instance = $serviceLocator->get($instance);
         }
 
-        switch ($this->name) {
-            case 'filesystem':
+        switch ($class) {
+            case 'Doctrine\Common\Cache\FilesystemCache':
                 $cache = new $class($options->getDirectory());
                 break;
-
-            case 'predis':
+            case 'DoctrineModule\Cache\ZendStorageCache':
+            case 'Doctrine\Common\Cache\PredisCache':
                 $cache = new $class($instance);
                 break;
-
             default:
                 $cache = new $class;
         }
--- a/tests/DoctrineModuleTest/Service/CacheFactoryTest.php
+++ b/tests/DoctrineModuleTest/Service/CacheFactoryTest.php
@@ -59,6 +59,7 @@ class CacheFactoryTest extends BaseTestCase
 
     /**
      * @covers \DoctrineModule\Service\CacheFactory::createService
+     * @group 547
      */
     public function testCreateZendCache()
     {
@@ -85,7 +86,8 @@ class CacheFactoryTest extends BaseTestCase
                     ]
                 ]
             ]
-        )->addAbstractFactory('Zend\Cache\Service\StorageCacheAbstractServiceFactory');
+        );
+        $serviceManager->addAbstractFactory('Zend\Cache\Service\StorageCacheAbstractServiceFactory');
 
         $cache = $factory->createService($serviceManager);
 
--- a/tests/DoctrineModuleTest/Service/CacheFactoryTest.php
+++ b/tests/DoctrineModuleTest/Service/CacheFactoryTest.php
@@ -57,6 +57,41 @@ class CacheFactoryTest extends BaseTestCase
         $this->assertSame('bar', $service->getNamespace());
     }
 
+    /**
+     * @covers \DoctrineModule\Service\CacheFactory::createService
+     */
+    public function testCreateZendCache()
+    {
+        $factory        = new CacheFactory('phpunit');
+        $serviceManager = new ServiceManager();
+        $serviceManager->setAlias('Config', 'Configuration');
+        $serviceManager->setService(
+            'Configuration',
+            [
+                'doctrine' => [
+                    'cache' => [
+                        'phpunit' => [
+                            'class' => 'DoctrineModule\Cache\ZendStorageCache',
+                            'instance' => 'my-zend-cache',
+                            'namespace' => 'DoctrineModule',
+                        ],
+                    ],
+                ],
+                'caches' => [
+                    'my-zend-cache' => [
+                        'adapter' => [
+                            'name' => 'blackhole'
+                        ]
+                    ]
+                ]
+            ]
+        )->addAbstractFactory('Zend\Cache\Service\StorageCacheAbstractServiceFactory');
+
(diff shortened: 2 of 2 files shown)


src/DoctrineModule/Service/CacheFactory.php
```diff
@@ -58,15 +58,14 @@ public function createService(ServiceLocatorInterface $serviceLocator)
             $instance = $serviceLocator->get($instance);
         }
 
-        switch ($this->name) {
-            case 'filesystem':
```
Ocramius: I initially thought that this would be a BC break, but it makes sense now: relying on checking against the class name makes more sense than relying on a very unsafe/unreliable `'filesystem'` string

src/DoctrineModule/Service/CacheFactory.php
```diff
@@ -58,15 +58,14 @@ public function createService(ServiceLocatorInterface $serviceLocator)
...
+            case 'Doctrine\Common\Cache\FilesystemCache':
                 $cache = new $class($options->getDirectory());
                 break;
-
-            case 'predis':
+            case 'DoctrineModule\Cache\ZendStorageCache':
+            case 'Doctrine\Common\Cache\PredisCache':
                 $cache = new $class($instance);
                 break;
-
```
Ocramius: To be reverted: please keep spacing between logical blocks

src/DoctrineModule/Service/CacheFactory.php
```diff
@@ -58,15 +58,14 @@ public function createService(ServiceLocatorInterface $serviceLocator)
             $instance = $serviceLocator->get($instance);
         }
 
-        switch ($this->name) {
-            case 'filesystem':
+        switch ($class) {
+            case 'Doctrine\Common\Cache\FilesystemCache':
                 $cache = new $class($options->getDirectory());
                 break;
-
```
Ocramius: To be reverted: please keep spacing between logical blocks

tests/DoctrineModuleTest/Service/CacheFactoryTest.php
```diff
@@ -57,6 +57,41 @@ public function testWillSetNamespace()
...
+                ],
+                'caches' => [
+                    'my-zend-cache' => [
+                        'adapter' => [
+                            'name' => 'blackhole'
+                        ]
+                    ]
+                ]
+            ]
+        )->addAbstractFactory('Zend\Cache\Service\StorageCacheAbstractServiceFactory');
```
Ocramius: Avoid reliance on the fluent interface here, please

tests/DoctrineModuleTest/Service/CacheFactoryTest.php
```diff
@@ -57,6 +57,41 @@ public function testWillSetNamespace()
         $this->assertSame('bar', $service->getNamespace());
     }
 
+    /**
+     * @covers \DoctrineModule\Service\CacheFactory::createService
```
Ocramius: Please add `@group 547`

BreiteSeite: I can do that, but shouldn't that a reference to an issue and not to a PR? (This PR doesn't have a related issue)

tests/DoctrineModuleTest/Service/CacheFactoryTest.php
BreiteSeite: Will update - i did not liked that either but didn't want to break "style" with the other test method that already existed (testCreatePredisCache).

tests/DoctrineModuleTest/Service/CacheFactoryTest.php
Ocramius: @BreiteSeite this PR is the issue (you explain the problem quite clearly as well)

BreiteSeite: Yes i know but do you use `@group <issue>` or `@group <pull request>`? Because mixing these both seems like a bad idea cause there could also be an issue with the ID 547 and that could confuse. Maybe something like `@group pr-547` would make clear the 547 isn't an issue id?

src/DoctrineModule/Service/CacheFactory.php
gianarb: I am agree.. but I have not tagged the 1.0 :) 
I can merge it and tag 1.0 with a BC Break..

BreiteSeite: This is only a very minor BC break. `new CacheFactory('predis')` works as before (not internally of course).
Only BC break would be if someone registered another class under the `predis` key and relied that the storage (`$instance`) would get injected into the constructor. Anything else works just like before. So i'm not sure if you want to release a new major version (technically/strictly it probably would be the right thing to do).

New major or not, i would appreciate a new release soon. :-) Thank you very much in advance.

tests/DoctrineModuleTest/Service/CacheFactoryTest.php
Ocramius: @BreiteSeite as long as it is clear where it comes from, this is fine :-)

PR or issue makes no big difference, as they have unique IDs anyway

BreiteSeite: ah. didn't knew github uses unique IDs for PRs and issues. Makes sense. Updated the PR anyways. :-) Thank you.

Ocramius: @BreiteSeite fyi, it would be impossible to reference things via `#number`, if they weren't unique ;-)

gianarb:
THanks a lot!

PR #549, opened 2016-01-31
PSR-4

TomHAnderson:
Up from PSR-0


Ocramius:
Please change only `composer.json`. Doctrine projects follow PSR-0 paths

gianarb:
I think that we can use PR-0 to stay in the same line of other doctrine's projects

TomHAnderson:
If Doctrine == PSR-0 than this PR isn't inline with standard.

PR #551, opened 2016-02-24
Update composer.json to allow for doctrine/common 2.6.*

basz:
I'm not sure if this makes sense or is dangerous. But I think it would solve this

https://github.com/zf-fr/zfr-oauth2-server-doctrine/issues/5

I currently do this in my project composer.json which can't be recommended either... :-)

"doctrine/common": "2.6.1 as 2.5.4"

PR #553, opened 2016-03-06
Drop php5.4

gianarb:
Reference with #551

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
         }
     ],
     "require": {
-        "php":                                ">=5.5",
+        "php":                                "^5.5 || ^7.0",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -31,8 +31,8 @@
         }
     ],
     "require": {
-        "php":                                ">=5.4",
-        "doctrine/common":                    ">=2.4,<2.6-dev",
+        "php":                                ">=5.5",
```
alcaeus: For future compatibility, this should be `^5.5 || ^7.0`. Using `>=` for PHP version constraints isn't the best idea (you can't really guarantee this will be compatible with PHP 8, can you? :wink:)

PR #560, opened 2016-04-17
Hydrator ignores private/protected getter/setter

PowerKiKi:
If the entity had a getter/setter which was private or protected the hydrator tried to call it anyway 
and thus threw an exception. We now silently ignore them if there are not callable.

This allow for entities that have "read-only" fields which are set internally (such as timestamp).

### DIFF
--- a/src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
+++ b/src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
@@ -258,7 +258,7 @@ class DoctrineObject extends AbstractHydrator
                 $target = $metadata->getAssociationTargetClass($field);
 
                 if ($metadata->isSingleValuedAssociation($field)) {
-                    if (! method_exists($object, $setter)) {
+                    if (! is_callable([$object, $setter])) {
                         continue;
                     }
 
@@ -275,7 +275,7 @@ class DoctrineObject extends AbstractHydrator
                     $this->toMany($object, $field, $target, $value);
                 }
             } else {
-                if (! method_exists($object, $setter)) {
+                if (! is_callable([$object, $setter])) {
                     continue;
                 }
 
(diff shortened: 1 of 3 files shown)


gianarb:
Thanks @Ocramius review before merge?

Ocramius:
Missing tests

PowerKiKi:
@Ocramius, thanks for not letting me sneak this patch without tests. Here there are, squashed in the original commit.

gianarb:
Thanks @PowerKiKi

BreyndotEchse:
If the class of `$object` implements the magic method `__call` then the condition `if (is_callable([$object, $setter])` would always return true. Wouldn't it be better to check both, if the method exists and is callable? `if (method_exists($object, $setter) && is_callable([$object, $setter])`

PowerKiKi:
It would mean that magic methods can never be used for hydrating. For me this would be OK, but this might be a breaking behavior for others. I suppose it should be debated in its own issue though, rather than on this merged PR.

Ocramius:
Can we simply mirror the behaviour of the ClassMethods hydrator in ZF?
Also, if this scenario is not covered there, we should escalate the
discussion to that component.

BreyndotEchse:
The behaviour of ClassMethods hydrator is pretty much the same. Same issue with `__call`

Ocramius:
The issue should be escalated there then.

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

izub:
@Ocramius coming back on this, the issue was escalated and a solution was implemented in zend-hydrator. See https://github.com/zendframework/zend-hydrator/pull/44
Would you agree it's a good idea to also apply that to the DoctrineObject hydrator here? As you mentioned it should mirror the behaviour of the ClassMethods hydrator.

PowerKiKi:
The potential change you are discussing is about __call method, and don't have much to do with the original intent of this PR of ignoring private/protected methods. IMHO you should create a new issue/PR for it.

izub:
@PowerKiKi Oh I agree. Just thought I'd quickly check if we want this before spending time on creating a new issue

PR #564, opened 2016-07-02
ZF3 support [BC break]

webimpress:
Prepared for release DoctrineModule v2 which will support only ZF3 on PHP 5.6 and 7.0.
In this release support of ZF2 should be dropped (it's very hard /or even impossible/ prepare package supports zend-mvc v2 and v3).

Library reorganized to use PSR-4 autoloading.

Documentation needs update.

Any suggestions would be appreciated.


snapshotpl:
Where is a problem with keep zend-mvc v2 support?

webimpress:
In zend-mvc v3 all console-related functionality is moved into `zend-mvc-console` and the package conflicted with zend-mvc v2:
https://github.com/zendframework/zend-mvc-console/blob/master/composer.json#L34
(Please notice also the zend-mvc-console package is not required by zend-mvc v3)
https://github.com/zendframework/zend-mvc/releases/tag/release-3.0.0

Unfortunately, there's no ability to do conditional requires in Composer, so we cannot specify a rule like "if zend-mvc v3 is installed, install zend-mvc-console as well".

manuakasam:
That has been a lot of work. Good job.

TomHAnderson:
I tried moving to PSR-4 but the Doctrine standard, I was told, is PSR-0.

When you're ready to merge this please post.  I'd like to try to make it work with Doctrine in Apigility.

webimpress:
@TomHAnderson I've moved back library to use PSR-0 autoloading instead of PSR-4.
I'm working also on changes in `DoctrineORMModule` for ZF3, probably tomorrow it will be done.
Let me know please if you find any issues with Apigility.

gianarb:
Yes @TomHAnderson is right.. All doctrine projects are using PSR-0 I think. It's like a standard :) @Ocramius can help about this information

webimpress:
@gianarb I've changed to use PSR-0, but I'm not sure if it's good, because PSR-0 is deprecated:

> Deprecated - As of 2014-10-21 PSR-0 has been marked as deprecated. PSR-4 is now recommended as an alternative.
http://www.php-fig.org/psr/psr-0/

Maybe will be nice then finally move to PSR-4? It's gonna be quite big update for the package... What do you think?

gianarb:
Yes I have not big problem with that but if there is some standard provided by doctrine the idea it's just follow that :) I will come back with some news about that.. and in the meantime I will review something here :) thanks for your effort @webimpress !

src/DoctrineModule/Module.php
```diff
@@ -49,18 +47,6 @@ class Module implements ConfigProviderInterface, InitProviderInterface, Bootstra
...
      */
-    public function init(ModuleManagerInterface $moduleManager)
-    {
-        AnnotationRegistry::registerLoader(
-            function ($className) {
-                return class_exists($className);
-            }
-        );
-    }
-
```
MatthiasKuehneEllerhold: Removing this function breaks the annotation driver:
[Doctrine\Common\Annotations\AnnotationException]
[Semantical Error] The annotation "@Doctrine\ORM\Mapping\Table" in class <CLASS> does not exist, or could not be auto-loaded.

webimpress: You have right, thanks. I've reverted this method and added some test to check if it works properly.
I don't know what is the plan about doctrine/annotations - maybe version 2.0 will be released before this, so then we can improve it (please see https://github.com/doctrine/annotations/pull/75)

webimpress:
@gianarb rebased and reverted to PSR-4 autoloading, if you decide to use anyway deprecated PSR-0 than maybe better if it will be done after merge or just before merge. I've started this PR from changing autoloading so it will be quite messy if we keep changing over and over ...

Saeven:
@webimpress thanks so much .  This opens up ZF3 for me, can't wait. @gianarb Thanks for expediting this!

TomHAnderson:
@webimpress I'm ready to move on this.  Will you look it over, since it's been some months so I can get your go-ahead on merging to the BC 2.0.0 branch?

webimpress:
No longer needed. DoctrineModule 1.2.0 supports ZF2 and ZF3. Please see https://github.com/doctrine/DoctrineModule/pull/567

PR #568, opened 2016-08-05
Changed isset for array_key_exists in context check

vincequeiroz:
**isset()** does not return TRUE for array keys that correspond to a NULL value, while **array_key_exists()** does.


gianarb:
Thanks @vincequeiroz

PR #569, opened 2016-08-17
Add missing Doctrine APCu Cache (PHP7)

JulienBreux:


gianarb:
Hello,
Why this change? We don't need to set nothing for cache like default value

Thanks

JulienBreux:
Hello @gianarb,

Thanks for your time and your Doctrine contribution.

Because it's here: https://github.com/doctrine/cache/tree/master/lib/Doctrine/Common/Cache

I think that's important to reflect cache classes in module configuration.

And because APC is deprecated because opcache.
I think that's interesting to show `Apcu` in cache classes.

gianarb:
Thanks a lot @JulienBreux

chris-fa:
The change in this PR was deleted in the next commit: https://github.com/doctrine/DoctrineModule/commit/6706e63632359dc43ac4745fc3b2246e0666cccc, I think it needs to be re-merged

PR #570, opened 2016-08-22
Expose to `zend-component-installer` as module `DoctrineModule`

webimpress:
There is new library used with zend application skeleton and apigility skeleton to simplify installation of new modules: [`zend-component-installer`](https://github.com/zendframework/zend-component-installer).


gianarb:
Thanks @webimpress

PR #571, opened 2016-08-23
ZF3 Composer dependencies - hotfix

webimpress:
Decreased composer dependencies of zend modules to match dependencies in ZendSkeletonApplication.
Currently if we build new project using composer `create-project` (from `ZendSkeletonApplication`) we have to run `composer update` because locked dependencies there are lower than required here.
We can safely decrees these dependencies in this library.


gianarb:
Thanks

PR #572, opened 2016-08-23
Command line tools improvements

webimpress:
`bin/doctrine-module.php` -> `bin/doctrine-module` - currently composer on installation creates proper files (bat/php files depends on system) in `vendor/bin` (so redundant files are now removed).

Now doctrine-module command line tool uses development configuration if it's enabled (the same way as it is done in zend skeleton application).


gianarb:
@Ocramius is it a BC break?

bin/doctrine-module
```diff
@@ -1,4 +1,63 @@
...
+} elseif (file_exists(__DIR__ . '/../vendor/autoload.php')) {
+    include_once __DIR__ . '/../vendor/autoload.php';
+} elseif (file_exists(__DIR__ . '/../../../autoload.php')) {
+    include_once __DIR__ . '/../../../autoload.php'; 
+} else {
+    throw new RuntimeException('Error: vendor/autoload.php could not be found. Did you run php composer.phar install?');
+}
+
+$appConfig = include 'config/application.config.php';
+if (file_exists('config/development.config.php')) {
```
Ocramius: This seems to be non-standard. Did I miss something? I never had different `development` configs.

bin/doctrine-module.bat
```diff
@@ -1,9 +0,0 @@
-@echo off
```
Ocramius: This file should at least be retained to prevent breaking existing apps relying on this path.

bin/doctrine-module.php
```diff
@@ -1,57 +0,0 @@
-<?php
```
Ocramius: This file should at least be retained to prevent breaking existing apps relying on this path.

bin/doctrine-module
webimpress: When you enable development mode in Zend `development.config.php.dist` is copied to `development.config.php`:
https://github.com/zendframework/ZendSkeletonApplication/tree/master/config

You can enable development modules in this file, disable cache etc...

bin/doctrine-module.bat
webimpress: This file shouldn't be used, I think it was there before only to use it in composer scripts. It's not required to have this `bat` file anymore because it will be automatically created in `vendor/bin` on windows.
I can agree it could be BC break for some applications, because we removed this file, but I think everywhere always is pointed to use vendor/bin/doctrine-module not files from `vendor/doctrine/DoctrineModule/bin/...`

webimpress:
@gianarb reverted `doctrine-module.php` and `doctrine-module.bat` to keep backward compatibility

gianarb:
You are right.. but we don't know :)

gianarb:
Thanks @webimpress for your big big effort!

PR #574, opened 2016-09-26
Removed direct interface implementations

webimpress:
These will be phased out of ModuleManager and pushed to the appropriate components in the near future.

Module Manager checks if interface is implemented or method exists.
This direct implementation of interfaces is removed in all ZF and Apigility modules so it should be removed also here for future compatibility. **No BC Break!**

### DIFF
--- a/src/DoctrineModule/Module.php
+++ b/src/DoctrineModule/Module.php
@@ -20,9 +20,6 @@
 namespace DoctrineModule;
 
 use Doctrine\Common\Annotations\AnnotationRegistry;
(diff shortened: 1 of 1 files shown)


Ocramius:
As you say, this is something that "will be phased out of ModuleManager and pushed to the appropriate components in the near future".

I don't want any duck-typing to land in here, so we should likely just wait for the new interfaces to pop up, and we then implement those.

Until then, removing the `implements` keyword is just a mistake that will lead to bigger issues.

PR #575, opened 2016-10-16
General code fixes

webimpress:
- updated PHP CodeSniffer to version 2.7
- added phpcs.xml with defined CS rules
- fixed code to pass all CS checks
- short array syntax in docs
- composer scripts to run unit and cs checks
- updated travis configuration to use composer scripts

ping @gianarb

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,9 +68,9 @@ install:
   - composer show
 
 script:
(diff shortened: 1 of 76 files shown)


TomHAnderson:
Please use hotfix/ prefix for branches instead of fix/

https://www.evernote.com/shard/s82/sh/2502ea74-1e39-4e6e-ac95-80145b948261/0bccfbd5032639372309e8577efe5e3b

PR #579, opened 2016-11-01
Added zend-mvc-console to require

TomHAnderson:


thomasvargiu:
I think you should not add zend-mvc-console, it's not compatible with zend-mvc ^2.0.
You should just suggest it.

```
    "conflict": {
        "zendframework/zend-mvc": "<3.0.0"
    },
```

TomHAnderson:
You'll see if you remove the composer.lock that the unit tests are
failing.  And it will be in the 2.0-dev branch anyhow.

TomHAnderson:
Thanks @thomasvargiu I see you're pointing out zend-mvc-console is deprecated.  I appreciate that.  I'll keep working on making the unit tests work before I move anything to a branch.

PR #581, opened 2016-11-17
Fix branch aliases. (For 2.0-dev)

MichaelGooden:
Branch aliases need to reference the actual branch names being used.

This fix is for the 2.0-dev branch.


TomHAnderson:
Hi Michael, the branch aliases I had in place would point a dev-master or dev-develop to the correct branches.  Are those branch aliases unnecessary and the aliases you gave here preferred?

PR #590, opened 2017-01-07
Added ConfigProvider in line with ZF components

asgrim:
Many ZF components now use a `ConfigProvider` to assist in config merging and allow components to be used e.g. in Expressive.

I took the strategy of least interference with existing config loading, so I did not move config here, nor make `Module.php` load config using the `ConfigProvider`. If desired, however, I can add this functionality into this PR. Feedback welcome, thanks!


asgrim:
Test failures don't look related to my change.

src/DoctrineModule/ConfigProvider.php
```diff
@@ -0,0 +1,41 @@
...
+{
+    /**
+     * @return array
+     */
+    public function __invoke()
+    {
+        $config = include __DIR__ . '/../../config/module.config.php';
+        $config['dependencies'] = $config['service_manager'];
+        unset($config['service_manager']);
+        return $config;
```
TomHAnderson: extra space before return statement

TomHAnderson:
The failing error has been corrected in master.  Please go ahead and put this into the Module.php as this is the direction ZF is now headed.

asgrim:
@TomHAnderson thanks for review; I've gone all-in with this now and moved all of `module.config.php` into the `ConfigProvider`. This would only break components which include `module.config.php` directly, but if you see that could be an issue, I could create a stub `module.config.php` again but use the `ConfigProvider` to return the config.

jaapio:
In order to add better support I think we should remove the zend-mvc from this package as well. Since expressive doesn't need that either. Should I open a new PR for that?

TomHAnderson:
This module needs a lot of work and that's coming up soon.  

We've got a lot of bugs here and while this module falls behind DoctrineORMModule is clean as a whistle.  Only recently have other Doctrine libraries fallen into line with Zend to use PHP 7.1 going forward.  

Moving this module towards Expressive is a goal and any PRs you can contribute, and indeed any guidance on the architecture to support these new libraries will be appreciated.

jaapio:
@TomHAnderson let me know if and how I can help to move forward to a beter module. If you create an project and issues that we can discuss I will try to contribute more to this module. We are using it on a daily base. So there must be time to help you guys move forward

PR #592, opened 2017-02-08
Add a params controller plugin

Erikvv:
Issue https://github.com/doctrine/DoctrineORMModule/issues/384

correspondig factory and module config is in https://github.com/doctrine/DoctrineORMModule/pull/522

### DIFF
--- a/src/DoctrineModule/Controller/Plugin/Params.php
+++ b/src/DoctrineModule/Controller/Plugin/Params.php
@@ -1,0 +1,135 @@
+<?php
+
+namespace DoctrineModule\Controller\Plugin;
+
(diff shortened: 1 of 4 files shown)


TomHAnderson:
I find this a little silly.  It looks like an idea you implemented and thought it would go well in this base repo.  I feel this idea is too specific a use for DoctrineModule.  It's clever but in the end it's just a shortcut that doesn't validate permissions making it possible to fetch any object by id without checking the id and it is this "skipped step" which I feel is a pot hole for other developers.  Props for putting it forward and it may find a home in it's own module but not this one.

PR #595, opened 2017-03-30
Update DoctrineObject.php

khaperets:


src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
```diff
@@ -58,9 +58,8 @@ class DoctrineObject extends AbstractHydrator
      */
     protected $byValue = true;
 
-
     /**
-     * Constructor
+     * Constructor.
```
TomHAnderson: The rest of this module does not end everything with a period.
https://github.com/doctrine/DoctrineModule/blob/e2ab9bbb9668196087cb7e83afb2b96a91d9c013/src/DoctrineModule/Stdlib/Hydrator/Strategy/AbstractCollectionStrategy.php#L117

This PR seems really superficial and if changes like periods are to be applied in this one class shouldn't they be applied across all Doctrine projects?

PR #599, opened 2017-04-27
Authentication guide update for ZF3

felzend:
Just a update for Zend Framework 3.

docs/Authentication.md

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -173,6 +173,38 @@ public function loginAction()
 }
 ```
 
+Instead of ZF2, you can do like this in ZF3:
+
+```php
+
+public function __construct($authenticationService)
+{
+    $this->authenticationService = $authenticationService;
+}
+
+public function loginAction()
+{
+    $data = $this->getRequest()->getPost();
+
+    // If you used another name for the authentication service, change it here
+    $authService = $this->authenticationService;
+
+    $adapter = $authService->getAdapter();
+    $adapter->setIdentity($data['login']);
+    $adapter->setCredential($data['password']);
+    $authResult = $authService->authenticate();
+
+    if ($authResult->isValid()) {
+        return $this->redirect()->toRoute('home');
+    }
+
+    return new ViewModel([
+        'error' => 'Your authentication credentials are not valid',
+    ]);
+}
+
+```
+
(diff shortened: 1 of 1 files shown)


docs/authentication.md
```diff
@@ -173,6 +173,38 @@ public function loginAction()
 }
 ```
 
+Instead of ZF2, you can do like this in ZF3:
```
TomHAnderson: Instead of Insteading ZF2, just create a section specific to ZF3

docs/authentication.md
```diff
@@ -173,6 +173,38 @@ public function loginAction()
...
+
+```php
+
+public function __construct($authenticationService)
```
TomHAnderson: Typecast the authentication service

docs/authentication.md
```diff
@@ -173,6 +173,38 @@ public function loginAction()
...
+
+public function loginAction()
+{
+    $data = $this->getRequest()->getPost();
+
+    // If you used another name for the authentication service, change it here
+    $authService = $this->authenticationService;
+
+    $adapter = $authService->getAdapter();
+    $adapter->setIdentity($data['login']);
```
TomHAnderson: use `$this->params()->fromPost('login')`

docs/authentication.md
```diff
@@ -173,6 +173,38 @@ public function loginAction()
...
+    $adapter->setCredential($data['password']);
+    $authResult = $authService->authenticate();
+
+    if ($authResult->isValid()) {
+        return $this->redirect()->toRoute('home');
```
TomHAnderson: ZF2 is `$this->plugin('redirect')->toRoute('home');`  Is this use of `redirect()` correct, a short hand?

docs/authentication.md
```diff
@@ -173,6 +173,38 @@ public function loginAction()
...
+    }
+
+    return new ViewModel([
+        'error' => 'Your authentication credentials are not valid',
```
TomHAnderson: Fetch the error from the `$authResult`.  Let it tell the problem; do not make one up.

docs/authentication.md
ddelrio1986: Yes this is a shorthand way of using the redirect plugin.

PR #604, opened 2017-09-01
Added factories for (No)ObjectExists and UniqueObject validators

fabiang:
Hello,

I've created factories for the following validators ObjectExist, NoObjectExists and UniqueObject validators. First this helps developers to create instance of those validators more easily. Secondly the validators are registered by default to the validator manager, so developers can use them straight away when using this module.

If this PR is merged you can use those validators right from your `getInputFilterSpecification()` method in your form/fieldset (as you would do this all other validators):

```php
<?php

use DoctrineModule\Validator\NoObjectExists as NoObjectExistsValidator;
use Zend\Form\Form;
use Zend\InputFilter\InputFilterProviderInterface;
use Doctrine\Common\Persistence\ObjectManager as DoctrineObjectManager;
use MyProject\Entity\Foobar as FoobarEntity;

class Myform extends Form implements InputFilterProviderInterface
{
    /**
     * @var DoctrineObjectManager
     */
    private $objectManager;

    public function __construct(DoctrineObjectManager $objectManager)
    {
        $this->objectManager = $objectManager;
    }

    public function getInputFilterSpecification()
    {
        return [
            'somefield' => [
                'required' => true,
                'validators' => [
                    [
                        'name'    => NoObjectExistsValidator::class,
                        'options' => [
                            'object_manager' => $this->objectManager,
                            'target_class'   => FoobarEntity::class,
                            'fields'         => ['somefield'],
                            'messages' => [
                                'objectFound' => 'An foo with this fieldname is already exists',
                            ],
                        ]
                    ]
                ]
            ]
        ];
    }
}
```

* All validators are aliased, so you can use them either by that alias or their FQCN:
  * DoctrineNoObjectExists => NoObjectExists
  * DoctrineObjectExists => ObjectExist
  * DoctrineUniqueObject => UniqueObject
* all options passed to the factories are merged an passed to the validator as options: this allows "messages" and so on to be changed
* this PR is compatible with ZF2 and 3
* this option keys have the same names like those when creating elements

I've got two questions:
1. I can provide tests. Should I use the old PHPUnit mocking framework or Prophecy?
2. Which branch I'm developing against? I see dev-1.0, dev-2.0 and master branches: what is the current develop version? Is the 1.0 branch only a maintain branch and new features are added to dev-2.0?

Thanks.


fabiang:
If everything is ok, I'll update the documentation to point out how to use those factories.

fabiang:
Added tests. Build seems not to fail for my changes.

src/DoctrineModule/Validator/Service/AbstractValidatorFactory.php
```diff
@@ -0,0 +1,135 @@
...
+            throw new ServiceCreationException(sprintf(
+                "Option 'target_class' is missing when creating validator %s",
+                __CLASS__
+            ));
+        }
+
+        $objectManager    = $this->getObjectManager($container, $options);
+        $targetClassName  = $options['target_class'];
+        $objectRepository = $objectManager->getRepository($targetClassName);
+        return $objectRepository;
```
TomHAnderson: Please leave a blank line before any return statement which is not the only line of code within brackets.

OK:
```
   $x = 1;

    return $x;
```

``` 
    if ($x) {
        return $x;
    }
```

Not ok:
```
    $x = 1; 
    return $x;
```

This should really be documented better.  I'll look into it.

src/DoctrineModule/Validator/Service/AbstractValidatorFactory.php
```diff
@@ -0,0 +1,135 @@
...
+ * Factory for creating NoObjectExists instances
+ *
+ * @license MIT
+ * @link    http://www.doctrine-project.org/
+ * @since   1.3.0
+ * @author  Fabian Grutschus <f.grutschus@lubyte.de>
+ */
+abstract class AbstractValidatorFactory implements FactoryInterface
+{
+    const DEFAULT_OBJECTMANAGER_KEY = 'Doctrine\ORM\EntityManager';
```
TomHAnderson: This is out of place.  Because this is DoctrineModule and not DoctrineORMModule we can't default to either ORM or ODM and be consistent.  

I agree ORM is the correct default here but we need to default to the service manager key for ORM which is `doctrine.entitymanager.orm_default` and not the EntityManager class name.

src/DoctrineModule/Validator/Service/Exception/ServiceCreationException.php
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * @license MIT
+ * @link    http://www.doctrine-project.org/
+ * @since   1.3.0
+ * @author  Fabian Grutschus <f.grutschus@lubyte.de>
+ */
+class ServiceCreationException extends BaseRuntimeException
+{
+
```
TomHAnderson: There's no reason for a blank line here.

TomHAnderson:
Hi @fabiang and thanks for the contribution.  This came in during the Labor Day holiday so please excuse the delay in getting back to you.  I've noted some issues here.  Some are your general coding style and I'll be happy to help you get these in line.

Please look though my notes and apply them globally.

fabiang:
Hello @TomHAnderson. Thanks for your review, I've applied the changes you suggested.

config/module.config.php
```diff
@@ -117,6 +119,19 @@
         ],
     ],
 
+    'validators' => [
```
TomHAnderson: Please rebase and add your configuration changes to the ConfigProvider.  This PR was merged ahead of your changes; thanks!

TomHAnderson:
I've got a project I can test with.  I'm still migrating the database from a different schema so it could be a few days but this is one PR I want to see added sooner than later.

fabiang:
Tested it within a Zend-Mvc project and it works.

TomHAnderson:
I've validated this with a Doctrine in Apigility application.

PR #605, opened 2017-09-08
Use service manager factory for cache when possible

bbatsche:
If a user has defined a factory for a given cache, we should use that factory to instantiate it. This will allow the user to define the logic and dependencies for their caches and give them greater flexibility when instantiating them.

### DIFF
--- a/src/DoctrineModule/Service/CacheFactory.php
+++ b/src/DoctrineModule/Service/CacheFactory.php
@@ -22,9 +22,8 @@ namespace DoctrineModule\Service;
 use Doctrine\Common\Cache\CacheProvider;
 use Interop\Container\ContainerInterface;
 use RuntimeException;
-use Doctrine\Common\Cache\MemcacheCache;
-use Doctrine\Common\Cache\MemcachedCache;
-use Doctrine\Common\Cache\RedisCache;
+use Doctrine\Common\Cache;
+use DoctrineModule\Cache\ZendStorageCache;
 use Zend\ServiceManager\ServiceLocatorInterface;
 
 /**
@@ -63,12 +62,12 @@ class CacheFactory extends AbstractFactory
             $cache = $container->get($class);
         } else {
             switch ($class) {
-                case \Doctrine\Common\Cache\FilesystemCache::class:
+                case Cache\FilesystemCache::class:
                     $cache = new $class($options->getDirectory());
                     break;
 
-                case \DoctrineModule\Cache\ZendStorageCache::class:
-                case \Doctrine\Common\Cache\PredisCache::class:
+                case ZendStorageCache::class:
+                case Cache\PredisCache::class:
                     $cache = new $class($instance);
                     break;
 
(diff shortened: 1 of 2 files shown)


src/DoctrineModule/Service/CacheFactory.php
```diff
@@ -59,18 +59,22 @@ public function __invoke(ContainerInterface $container, $requestedName, array $o
...
 
-        switch ($class) {
-            case 'Doctrine\Common\Cache\FilesystemCache':
-                $cache = new $class($options->getDirectory());
-                break;
+        if ($container->has($class)) {
+            $cache = $container->get($class);
+        } else {
+            switch ($class) {
+                case 'Doctrine\Common\Cache\FilesystemCache':
```
TomHAnderson: The new way to specify class names is to use `::class`.  Please change this to 
```
switch ($class) {
    case Cache\FilesystemCache::class: 
    ...
```

src/DoctrineModule/Service/CacheFactory.php
```diff
@@ -59,18 +59,22 @@ public function __invoke(ContainerInterface $container, $requestedName, array $o
...
-                break;
+                case 'DoctrineModule\Cache\ZendStorageCache':
+                case 'Doctrine\Common\Cache\PredisCache':
+                    $cache = new $class($instance);
+                    break;
 
-            default:
-                $cache = new $class;
+                default:
+                    $cache = new $class;
```
TomHAnderson: Even the default case needs a break statement:
http://www.php-fig.org/psr/psr-2/#52-switch-case

tests/DoctrineModuleTest/Service/CacheFactoryTest.php
```diff
@@ -120,4 +120,32 @@ public function testCreatePredisCache()
...
+    {
+        $factory        = new CacheFactory('chain');
+        $serviceManager = new ServiceManager();
+        $serviceManager->setService(
+            'Configuration',
+            [
+                'doctrine' => [
+                    'cache' => [
+                        'chain' => [
+                            'class' => 'Doctrine\Common\Cache\ChainCache',
```
TomHAnderson: use `::class` here too

src/DoctrineModule/Service/CacheFactory.php
```diff
@@ -63,17 +63,18 @@ public function __invoke(ContainerInterface $container, $requestedName, array $o
             $cache = $container->get($class);
         } else {
             switch ($class) {
-                case 'Doctrine\Common\Cache\FilesystemCache':
+                case \Doctrine\Common\Cache\FilesystemCache::class:
```
TomHAnderson: Classes should never be directly referenced.  This includes classes like Exception.  Please `use` these namespaces.  

For instance, for the `Doctrine\Common\Cache` namespace add `use Doctrine\Common\Cache as CommonCache;` then in code use `case CommonCache\FileSystemCache::class:`  This allows some level of creativity and the design extends from namespaces with lots of classes such as Doctrine entities where `use Db\Entity;` allows for `if ($entity instananceof Entity\User)`...

TomHAnderson: Since only one class is outside the Doctrine\Common\Cache namespace just using `use Doctrine\Common\Cache;` then `case Cache\FileSystemCache...` is nice and clean.

TomHAnderson:
@bbatsche I'll work around this next issue but it's important for you going forward.  For this PR you opened it against your own master branch.  The proper procedure is to create a new branch from master and apply your commits to that branch then open a PR for your branch. 

The rule of thumb is your master branch should always be able to be equal to the upstream (this) master branch.  If I merged a different PR before yours then your master branch would become out of sync.

bbatsche:
@TomHAnderson Thanks for the feedback. I'd be happy to move this over to its own branch and create a new pull request if that would help, or if it could get this merged any sooner.

PR #612, opened 2017-11-18
Allow installing Symfony 4

alcaeus:
This also adds a build against dev dependencies to test upcoming versions

#SymfonyConHackday2017


alcaeus:
Note: we might want to wait for #611 to be merged before merging this one. What do you think @TomHAnderson?

PR #617, opened 2018-01-03
Removing the license header

mikeSimonson:


ddelrio1986:
Why is the header being removed?

TomHAnderson:
@ddelrio1986 see 
https://github.com/doctrine/DoctrineORMModule/pull/547#issuecomment-355201097

I haven't investigated this yet.

mikeSimonson:
Because
- it's useless from a legal point of view
- half the file have it
- the other half doesn't
- the third half has various different version of it

We end up with only one license file at the root of the repo without any confusion possible because some file have various version of the header.

PR #622, opened 2018-03-06
Make Hydrator\DoctrineObject::handleTypeConversions handle all (simple) defaults 

XCame:
Extend DoctrineModule\Stdlib\Hydrator\DoctrineObject::handleTypeConversions() to handle all basic conversions according to documentation instead of just DateTime object related values

Stumbled across this, because I hadn't typecast some bools in an Entity setter method. That caused the UoW to detect changes e.g ` true === 1` and always updated the Entity although nothing changed (logically).

There are default type mappings from Doctrine to PHP via the hydrator, but not vice versa.
(see http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/basic-mapping.html#doctrine-mapping-types)

Now there are (at least the simple ones).

Before:
> Column definition is **boolean**
> $value in DB is **1** (TINYINT)
> $value extracted from hydrator is **true** _<== this should be the same in both directions_
> $value used in Form becomes **1**
> $value given to hydrator is **1**
> $value hydrated is **1** _<== but it isn't_

Now:
> Column definition is **boolean**
> $value in DB is **1** (TINYINT)
> $value extracted from hydrator is **true** _<== this should be the same in both directions_
> $value used in Form becomes **1**
> $value given to hydrator is **1**
> $value hydrated is **true** _<== now it is_

### DIFF
--- a/src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
+++ b/src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
@@ -466,7 +466,9 @@ class DoctrineObject extends AbstractHydrator
 
     /**
      * Handle various type conversions that should be supported natively by Doctrine (like DateTime)
-     *
+     * See Documentation of Doctrine Mapping Types for defaults
+     * 
+     * @link http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/basic-mapping.html#doctrine-mapping-types
      * @param  mixed  $value
      * @param  string $typeOfField
      * @return DateTime
@@ -474,6 +476,23 @@ class DoctrineObject extends AbstractHydrator
     protected function handleTypeConversions($value, $typeOfField)
     {
         switch ($typeOfField) {
+            case 'boolean':
+                $value = (bool)$value;
+                break;
+            case 'string':
+            case 'text':
+            case 'bigint':
+            case 'decimal':
+            case 'decimal':
+                $value = (string)$value;
+                break;
+            case 'integer':
+            case 'smallint':
+                $value = (int)$value;
+                break;
+            case 'float':
+                $value = (double)$value;
+                break;
             case 'datetimetz':
             case 'datetime':
             case 'time':


TomHAnderson:
Nice work!  I'm a fan of this kind of typecasting.  Can you provide a unit test to show that true === true and not === 1?

src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
```diff
@@ -466,14 +466,33 @@ function ($item) {
...
+     * 
+     * @link http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/basic-mapping.html#doctrine-mapping-types
      * @param  mixed  $value
      * @param  string $typeOfField
      * @return DateTime
      */
     protected function handleTypeConversions($value, $typeOfField)
     {
         switch ($typeOfField) {
+            case 'boolean':
```
Ocramius: Argh, why are we doing the conversion here in first place? This should be handled in DBAL types

XCame:
Besides the fact that I'm anything but good at setting up unit tests, I can try to provide some. 
But I think it won't happen until next month. I'm kind of overbooked at the moment.

Ocramius:
@XCame this applies to almost everyone working on OSS, we included. Simply send over the tests when you have a chance.

TomHAnderson:
@XCame Thanks.  Once we have tests for this we'll be very close to a 2.0 release.

src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
TomHAnderson: @Ocramius do you want to do anything about this or let it go ahead inside this module?  I'm ready to start writing unit tests to we can get a 2.1 out the door and let doctrine-orm-module rely on it.

Ocramius: Nothing to be fixed in this patch, but we'll have to consider this legacy to be removed...

PR #624, opened 2018-03-17
Use dedicated PHPUnit assertions

carusogabriel:


TomHAnderson:
Good changes, thanks!  I'll put them into a 2.0 release in the near future.

Issue #632, opened 2018-04-05
UniqueObject doesn't work anymore after upgrade when adding new entites


BigMichi1:
today we upgraded from 1.2.0 to 2.1.1 and we are now seeing an issue with the UniqueObject validator that worked before (1.2.0)

we have this in our fieldset
```
                    [
                        'name' => UniqueObject::class,
                        'options' => [
                            'object_manager' => $this->objectManager,
                            'object_repository' => $this->objectManager->getRepository(Operationtype::class),
                            'use_context' => true,
                            'fields' => [
                                'shortname'
                            ],
                            'messages' => [
                                'objectNotUnique' => 'message operationtype already exists with this shortname',
                            ],
                        ],
                    ]
```
this will now fail when we are adding a new entry because of the change in the method getExpectedIdentifiers() from ```!array_key_exists($identifierField, $context)``` to 
```! isset($context[$identifierField])``` and we got an exception about the missing identifier but it is in the array but null as it is correct for new entries, changing this one back to the old check everthing is now fine

TomHAnderson:
Because you have a change which works will you push it in a PR?

TomHAnderson:
@BigMichi1 I'd like to put a unit test in to ensure this doesn't happen again.  Can you provide the `$context` which breaks with this change?

PR #633, opened 2018-04-05
Revert to array_key_exists instead of isset

TomHAnderson:
https://github.com/doctrine/DoctrineModule/issues/632


Ocramius:
This requires a test case to be added

PR #636, opened 2018-05-02
Throw exception for null objects in DoctrineObject hydrator.

ariddlestone:
When calling $hydrator->extract(null) or $hydrator->hydrate(null), errors thrown are very unhelpful, as the ObjectManager tries to get the metadata for the hydrator class itself (as get_class(NULL) returns the name of the class it is called in).
I've just added a check that we are trying to prepare an object.


ariddlestone:
Sorry, I knew I'd forgotten something. Do you want me to submit this one again, or can you work with it as it is?

TomHAnderson:
The hydrator has been moved to https://github.com/doctrine/doctrine-laminas-hydrator

PR #643, opened 2018-08-02
The Inflector class is deprecated

nepda:
I've simply changed the `use` statement.

The Inflector class is deprecated, use Doctrine\Common\Inflector\Inflector from doctrine/inflector package instead.


Erikvv:
please tag this fix

PR #646, opened 2018-09-08
Apply hydrator strategies before type conversion

### LINKED ISSUE #639: BC break for hydrator strategy.
Demonstrated in #638

After https://github.com/doctrine/DoctrineModule/commit/23239885250c3c11458102ee954e33d87d3fad7d#diff-d845165302f39b8f7ff69f0652ae266d in #626 

values are parsed before strategies have a chance to do their conversion. Causes problems if source data type is not same as target.

Erikvv:
fixes #639

also:
- Renamed test asset SimpleEntity to ByValueDifferentiatorEntity and created a new SimpleEntity which is actually simple, to simplify the test without resorting to the less often used byReference hydration
- Changes in ->testObjectIsPassedForContextToStrategies so it still tests what it's supposed to. This test was affected by the same type conversions which caused the regression.


Erikvv:
I dunno if it's worth to add a separate test for byReference.
I checked to see if it works just to be sure.

Erikvv:
Please don't let one depend on the other. This breakage is not related to the implementation of handleTypeConversions, just the order in which it is called.

TomHAnderson:
These PRs will not depend on each other.  Because you're active with an eye on hydrators I'd like your opinion on the other one too.

PR #654, opened 2018-10-17
Revert "Check nullable fields on handleTypeConversions method"

TomHAnderson:
Reverts doctrine/DoctrineModule#651


floriancoding:
It wasn't me who submitted the PR but @vincequeiroz 

But I think it's not that easy, because in order to check if association fields are nullable we would need a new function in doctrine/orm. Current isNullable method does not decide between primitive and association types and therefore treats association fields as primitive fields, which is in return causing the MappingException.

PR #655, opened 2018-10-19
Type convertion bug fix for nullable fields

hkulekci:
I took the @vincequeiroz commit and added associated field support for #649

ping @TomHAnderson

### DIFF
--- a/src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
+++ b/src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
@@ -272,6 +272,10 @@ class DoctrineObject extends AbstractHydrator
     {
         $value = parent::hydrateValue($name, $value, $data);
 
+        if (is_null($value) && method_exists($this->metadata, 'isNullable') && $this->metadata->isNullable($name)) {
+            return null;
+        }
+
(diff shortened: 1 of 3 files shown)


hkulekci:
Getting the fail on the test because in some test cases there is a check for calling count of hasAssociation. One of them is below: 

```
public function testCanHydrateOneToOneAssociationByReferenceWithNullableRelation()
    {
        // When using hydration by reference, it won't use the public API of the entity to retrieve values (setters)
        $entity = new Asset\OneToOneEntity();

        $this->configureObjectManagerForOneToOneEntity();
        $this->objectManager->expects($this->never())->method('find');
        $this->metadata->expects($this->once())->method('hasAssociation');

        $data = ['toOne' => null];

        $object = $this->hydratorByReference->hydrate($data, $entity);
        $this->assertNull($object->getToOne(false));
    }
```

Why do we need only once calling check? Why should `hasAssociation` method be called only once?

```
$this->metadata->expects($this->atLeastOnce())->method('hasAssociation');
```

Because after changing `hydrateByValue` method with `$this->isNullable()` condition, `hasAssociation` method is calling two time in these tests.

hkulekci:
@TomHAnderson @vincequeiroz  I updated the tests part.

PR #662, opened 2019-01-22
Removing the license header

webimpress:
Continuation of b1e0f5b7de4d35e976468e5c49feabe32a852c23


ddelrio1986:
Why are the license headers being removed?

webimpress:
This is way we are doing it now in doctrine - the recent trend. We have LICENSE file in the project and it's sufficient. No need to duplicate it in every file.  See for example: https://github.com/doctrine/orm/pull/6600

PR #644, opened 2018-08-03
Extract hydrators to separate repository

michalbundyra:
I've created repository with the hydrators:
https://github.com/webimpress/doctrine-zend-hydrator

I want to use Doctrine with ZF Expressive application and I need hydrators there. `DoctrineModule` is specific for MVC ZF applications, and I don't want to install all of those dependencies just to use hydrators. I think many other parts we can extract to smaller packages, and make them reusable in Expressive applications.

I am happy to transfer this repository to doctrine project. We can change namespace etc. I didn't publish any release there. I just wanted to create something to start with :)

PR #663, opened 2019-01-25
Fixed hydration with nested entities

webimpress:
In case we try hydrate entity with nested entities metadata was overridden in the hydrator and invalid metadata was used to hydrate rest of the fields after hydrating nested entity.

The issue was introduced in [version 2.1.4](https://github.com/doctrine/DoctrineModule/releases/tag/2.1.4) in commit: 336b33ffdbddc9431f776646e5353ce0a5a6bcfb
PR: https://github.com/doctrine/DoctrineModule/pull/646

The tests are not the best in the library. I hope we will be able to improve them when we extract hydrators to separate repository. I would like to get class metadata for the entity instead of mocking it.
See: #644 

/cc @Erikvv @alextech @TomHAnderson

### DIFF
--- a/tests/DoctrineModuleTest/Stdlib/Hydrator/DoctrineObjectTest.php
+++ b/tests/DoctrineModuleTest/Stdlib/Hydrator/DoctrineObjectTest.php
@@ -2,10 +2,14 @@
 
 namespace DoctrineModuleTest\Stdlib\Hydrator;
 
+use Doctrine\Common\Persistence\Mapping\ClassMetadata;
+use Doctrine\Common\Persistence\ObjectManager;
 use DoctrineModuleTest\Stdlib\Hydrator\Asset\ByValueDifferentiatorEntity;
 use DoctrineModuleTest\Stdlib\Hydrator\Asset\ContextStrategy;
 use DoctrineModuleTest\Stdlib\Hydrator\Asset\SimpleEntity;
 use PHPUnit\Framework\TestCase as BaseTestCase;
+use Prophecy\Argument;
+use Prophecy\Prophecy\ObjectProphecy;
 use ReflectionClass;
 use Doctrine\Common\Collections\ArrayCollection;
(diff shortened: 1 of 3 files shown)


tests/DoctrineModuleTest/Stdlib/Hydrator/DoctrineObjectTest.php
```diff
@@ -2668,4 +2673,115 @@ public function hydrate($value) : string
 
         $this->assertEquals('complex,value', $entity->getField());
     }
+
+    public function testStrategyWithArrayByReference()
```
webimpress: Added test for #639 to cover also hydration by reference.

tests/DoctrineModuleTest/Stdlib/Hydrator/DoctrineObjectTest.php
```diff
@@ -2668,4 +2673,115 @@ public function hydrate($value) : string
...
+            }
+
+        });
+
+        $this->hydratorByReference->hydrate($data, $entity);
+
+        $this->assertSame('complex,value', $entity->getField());
+    }
+
+    private function getObjectManagerForNestedHydration()
```
webimpress: I know, this is not the nicest way, we would need some method to get class metadata for the entity, and I guess it shouldn't be mocked but generated based on the entity. I hope we can change it later on, as said in the PR description.

Erikvv:
I would prefer to instantiate another hydrator for the nested entity so we can avoid these state issues.

webimpress:
@Erikvv 

> I would prefer to instantiate another hydrator for the nested entity so we can avoid these state issues.

Yeah, maybe. But it was not done that way and we don't really know what impact it could have. I don't like that solution, as it is right now, but I believe we can improve later on - in separate repository with better tests. Please remember that this is a hotfix of BC Break introduced in version 2.1.4.

PR #664, opened 2019-02-13
bind Closure to null to allow the garbage collector do its work

rieschl:
The closure did bind to `$this` or `self` and therefore couldn't be garbage collected.

### DIFF
--- a/src/DoctrineModule/Module.php
+++ b/src/DoctrineModule/Module.php
@@ -35,7 +35,7 @@ class Module implements ConfigProviderInterface, InitProviderInterface, Bootstra
     public function init(ModuleManagerInterface $moduleManager)
     {
         AnnotationRegistry::registerLoader(
-            function ($className) {
+            static function ($className) {
                 return class_exists($className);
             }
         );


rieschl:
Credit goes to @Xerkus for discovering it 😄

rieschl:
To write a test for it we would need [Weakref](https://pecl.php.net/package/Weakref), which is probably not possible.

src/DoctrineModule/Module.php
```diff
@@ -35,9 +35,9 @@ class Module implements ConfigProviderInterface, InitProviderInterface, Bootstra
     public function init(ModuleManagerInterface $moduleManager)
     {
         AnnotationRegistry::registerLoader(
-            function ($className) {
+            (function ($className) {
```
Ocramius: You can use `static function ($className) {` (yep, that works ;-) )

Ocramius:
> To write a test for it we would need [Weakref](https://pecl.php.net/package/Weakref), which is probably not possible.

A test case would use `memory_get_usage()`, IMO

rieschl:
> A test case would use `memory_get_usage()`, IMO

I changed the fix (and squashed the commit) to use `static` but I'm not sure if I understand PHP internals good enough to write a meaningful test using `memory_get_usage()`.

rieschl:
Any chance someone could write a test for that so it can be merged?

webimpress:
@rieschl 
I was looking on it and really I don't know how it should work... I came up with:

```
    public function testInit()
    {
        $moduleManager = $this->prophesize(ModuleManager::class);

        echo PHP_EOL;
        echo memory_get_usage(), PHP_EOL;
        $module = new Module();
        for ($i = 0; $i < 10000; ++$i) {
            $module->init($moduleManager->reveal());
        }
        echo memory_get_usage(), PHP_EOL;
        AnnotationRegistry::reset();
        echo memory_get_usage(), PHP_EOL;
    }
```

but with/without your changes results are as I would expect:
- first the memory usage is the lowest
- after binding many loaders is the highest
- after clearing is lower than middle, but greater than first

@Ocramius would you be able to help, please?

TomHAnderson:
I'm ready to merge this unless someone wants to raise issue as-is.

PR #665, opened 2019-04-10
2 Bugfixes regarding recursive hydration

boesing:
Hey guys,

I've been using your module for about 3 years now.
We are recently experiencing some issues as we are trying to use the hydrator for recursive entity creations.

The first thing is: if one is using recursive hydration for `toMany` relations, the `toMany` method within the `DoctrineObject`-Hydrator is calling `DoctrineObject::hydrate` which overwrites the `metadata` property in its `prepare` method.

I've wrote a unit test for that and fixed the issue.

After this, I went to solve our current issue with the recursive hydration in `toMany` relations:
If there are relations having primary keys and one is providing just a partial set of the identifiers, `doctrine/orm` for example throws an `ORMException`:
```php
foreach ($class->identifier as $identifier) {
    if ( ! isset($id[$identifier])) {
        throw ORMException::missingIdentifierField($class->name, $identifier);
    }

    $sortedId[$identifier] = $id[$identifier];
    unset($id[$identifier]);
}
```

If the partial set of the identifiers contains all other identifiers but the missing identifier is, for example, the reference back to an entity, it should get properly created.

Here is an example:

```php
class EntityWithToMany
{
    public $id;
    /** @var EntityWithinToMany[]|Collection */
    public $toMany;

    public function __construct()
    {
        $this->toMany = new ArrayCollection();
    }
}

class EntityWithinToMany
{
    /** @var EntityWithToMany */
    public $toOne;

    /** @var integer */
    public $whateverOtherIdentifier;
}
```

If you now want to create an recursive entity like this, it would currently throw the exception:
```php
$entity = new EntityWithToMany();
$data = [
    'toMany' => [
        [
            // ofc I cannot provide the `toOne` identifier as we are going to create it
            'whateverOtherIdentifier' => 123,
        ],
    ],
];
$hydrator->hydrate($data, $entity); // in doctrine/orm this throws ORMException::missingIdentifierField
```

With this PR, the `$entity` information is being filled to the `value` information, if the information is missing.

- #663 fixed the `toOne` handling, this PR fixes the `toMany` handling.
- It introduces a feature of recursive hydration aswell.

If I have to strip the feature out into a different branch, please let me know.
I would love having that fix kinda asap as I am actually stucking in implementing a feature. Hopefully one gets time to review this.

Thanks in advance.

/cc @TomHAnderson

### DIFF
--- a/src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
+++ b/src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
@@ -3,15 +3,15 @@
 namespace DoctrineModule\Stdlib\Hydrator;
 
 use DateTime;
+use Doctrine\Common\Inflector\Inflector;
 use Doctrine\Common\Persistence\Mapping\ClassMetadata;
 use Doctrine\Common\Persistence\ObjectManager;
-use Doctrine\Common\Inflector\Inflector;
 use InvalidArgumentException;
 use RuntimeException;
 use Traversable;
-use Zend\Stdlib\ArrayUtils;
 use Zend\Hydrator\AbstractHydrator;
 use Zend\Hydrator\Filter\FilterProviderInterface;
+use Zend\Stdlib\ArrayUtils;
 
 /**
  * This hydrator has been completely refactored for DoctrineModule 0.7.0. It provides an easy and powerful way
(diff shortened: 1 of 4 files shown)


ddelrio1986:
@boesing why did you decide to close this and delete it?

boesing:
@ddelrio1986 I've tested my changes in our environment and realized that it does not work as expected.
I'll probably create another fix someday but actually I'll using a workaround.

PR #667, opened 2019-08-20
Support type conversion for the immutable field type variants in the Hydrator

netiul:
Support type conversion for the immutable field type variants to the Hydrator: datetime_immutable, datetimetz_immutable, date_immutable, time_immutable.

Should be BC proof.

### DIFF
--- a/src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
+++ b/src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
@@ -3,6 +3,7 @@
 namespace DoctrineModule\Stdlib\Hydrator;
 
 use DateTime;
+use DateTimeImmutable;
 use Doctrine\Common\Persistence\Mapping\ClassMetadata;
 use Doctrine\Common\Persistence\ObjectManager;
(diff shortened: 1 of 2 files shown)


src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php
```diff
@@ -565,25 +566,35 @@ protected function handleTypeConversions($value, $typeOfField)
...
                 break;
             case 'datetimetz':
+            case 'datetimetz_immutable':
             case 'datetime':
+            case 'datetime_immutable':
             case 'time':
+            case 'time_immutable':
             case 'date':
+            case 'date_immutable':
+                if (strpos($typeOfField, '_immutable', 0) !== false) {
```
SenseException: If the immutable and regular date/time cases can be separated, you can use a private function for creating the `DateTimeInterface` instance doing something like `private function handleDateTime(DateTimeImmutable::class, $value) : DateTimeInterface` or injecting a closure for instance creation instead of injecting the classname string.

TomHAnderson:
The hydrator has been moved to https://github.com/doctrine/doctrine-laminas-hydrator

PR #672, opened 2020-01-07
Test against PHP 7.4

Slamdunk:


.travis.yml
```diff
@@ -33,11 +33,9 @@ matrix:
     - php: 7.3
       env:
         - DEPS=latest
-    - php: nightly
```
greg0ire: Why should we stop testing against nightly?

Slamdunk: For what I understand reading the history, `nightly` was something used to represent the **next** PHP version. This was a good intention, but at the end was Travis builds as `nightly` has very little in common with whatever lands as a final release. So building against nightly is only a waste of electricity, and no one reads its log because it gives zero value: nightly errors may disappears once new PHP version is published.

greg0ire: I'm not against that decision. Please move it to a separate commit (not necessary a separate PR), and make sure to include this explanation in your commit message

Slamdunk: Extended explanation added in extended commit message, in a separate commit.

greg0ire:
Thanks @Slamdunk !

PR #683, opened 2020-02-16
Feature/coding standard

TomHAnderson:
This PR upgrades phpunit and phpcs and adds the Doctrine Coding Standard.

phpcs passes with the Doctrine standard.  

In making the many changes to make phpcs pass everything that was 
```
/**
 * @param array $something
```

became
```
/**
 * @param mixed[] $something
```

To research the origin of data for each array would be over taxing.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -59,7 +59,7 @@
         "laminas/laminas-servicemanager": "^3.3",
         "laminas/laminas-stdlib": "^3.2.1",
         "laminas/laminas-validator": "^2.10",
-        "symfony/console": "^3.3 || ^4.0 || ^5.0"
+        "symfony/console": "^5.0"
     },
     "require-dev": {
         "doctrine/coding-standard": "^7.0",
--- a/composer.json
+++ b/composer.json
@@ -95,7 +95,7 @@
             "@cs-check",
             "@test"
         ],
-        "cs-check": "phpcs",
+        "cs-check": "phpcs --standard=Doctrine src/",
         "cs-fix": "phpcbf",
         "test": "phpunit --colors=always",
         "test-coverage": "phpunit --coverage-clover build/clover.xml"
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,7 @@
         },
         "branch-alias": {
             "dev-1.2-dev": "1.2-dev",
-            "dev-2.0-dev": "2.0-dev",
+            "dev-2.0-dev": "2.0-dev"
         }
     },
     "require": {
--- a/composer.json
+++ b/composer.json
@@ -71,9 +71,9 @@
         "laminas/laminas-serializer": "^2.8",
         "laminas/laminas-session": "^2.8",
         "laminas/laminas-test": "^3.1.1",
-        "phpunit/phpunit": "^7.5.2",
+        "phpunit/phpunit": "^8.5",
         "predis/predis": "^1.1",
-        "squizlabs/php_codesniffer": "^2.7"
+        "doctrine/coding-standard": "^7.0"
     },
     "suggest": {
         "doctrine/data-fixtures":         "Data Fixtures if you want to generate test data or bootstrap data for your deployments",
--- a/composer.json
+++ b/composer.json
@@ -43,14 +43,13 @@
         "branch-alias": {
             "dev-1.2-dev": "1.2-dev",
             "dev-2.0-dev": "2.0-dev",
-            "dev-master": "1.2-dev",
-            "dev-develop": "2.0-dev"
         }
     },
     "require": {
         "php": "^7.2",
-        "doctrine/common": "^2.8",
         "doctrine/cache": "^1.7",
+        "doctrine/common": "^2.8",
+        "doctrine/doctrine-laminas-hydrator": "^2.0.2",
         "laminas/laminas-authentication": "^2.5.3",
         "laminas/laminas-cache": "^2.7.1",
         "laminas/laminas-form": "^2.11",
@@ -60,10 +59,10 @@
         "laminas/laminas-servicemanager": "^3.3",
         "laminas/laminas-stdlib": "^3.2.1",
         "laminas/laminas-validator": "^2.10",
-        "symfony/console": "^3.3 || ^4.0 || ^5.0",
-        "doctrine/doctrine-laminas-hydrator": "^2.0.2"
+        "symfony/console": "^3.3 || ^4.0 || ^5.0"
     },
     "require-dev": {
+        "doctrine/coding-standard": "^7.0",
         "laminas/laminas-i18n": "^2.7",
         "laminas/laminas-log": "^2.9",
         "laminas/laminas-modulemanager": "^2.8",
@@ -72,8 +71,7 @@
         "laminas/laminas-session": "^2.8",
         "laminas/laminas-test": "^3.1.1",
         "phpunit/phpunit": "^8.5",
-        "predis/predis": "^1.1",
-        "doctrine/coding-standard": "^7.0"
+        "predis/predis": "^1.1"
     },
     "suggest": {
         "doctrine/data-fixtures":         "Data Fixtures if you want to generate test data or bootstrap data for your deployments",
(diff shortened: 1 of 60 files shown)


composer.json
```diff
@@ -71,9 +71,9 @@
         "laminas/laminas-serializer": "^2.8",
         "laminas/laminas-session": "^2.8",
         "laminas/laminas-test": "^3.1.1",
-        "phpunit/phpunit": "^7.5.2",
+        "phpunit/phpunit": "^8.5",
         "predis/predis": "^1.1",
-        "squizlabs/php_codesniffer": "^2.7"
+        "doctrine/coding-standard": "^7.0"
```
greg0ire: Let's take this as a occasion to sort packages? Example: https://github.com/doctrine/orm/blob/e77091399a4475630a833fc33bff0d54c194511a/composer.json#L33-L35

src/DoctrineModule/Authentication/Adapter/ObjectRepository.php
```diff
@@ -140,18 +132,20 @@ protected function validateIdentity($identity)
      * This method abstracts the steps involved with making sure that this adapter was
      * indeed setup properly with all required pieces of information.
      *
-     * @throws Exception\RuntimeException - in the event that setup was not done properly
```
greg0ire: What was wrong with this? it's good to know which exception is thrown exactly, this was the way to do it. Maybe the issue you had was just about formatting?

src/DoctrineModule/ConfigProvider.php
```diff
@@ -30,64 +30,55 @@ public function __invoke()
     /**
      * Return application-level dependency configuration
      *
-     * @return array
+     * @return mixed[]
      */
-    public function getDependencyConfig()
+    public function getDependencyConfig() : array
     {
+// phpcs:disable
```
greg0ire: Why do we need this?

src/DoctrineModule/ServiceFactory/AbstractDoctrineServiceFactory.php
```diff
@@ -96,18 +98,19 @@ private function getFactoryMapping(ContainerInterface $serviceLocator, $name)
...
-            return [
-                'serviceType'  => $serviceType,
-                'serviceName'  => $serviceName,
-                'factoryClass' => $config['doctrine_factories'][$mappingType][$serviceType],
-            ];
         }
+
+        if (! isset($config['doctrine_factories'][$mappingType]) ||
+             ! isset($config['doctrine_factories'][$mappingType][$serviceType]) ||
+             ! isset($config['doctrine'][$mappingType][$serviceType][$serviceName])
```
greg0ire: Let's maybe change this to
```
!isset(
    $config['doctrine_factories'][$mappingType],
    $config['doctrine_factories'][$mappingType][$serviceType],
    $config['doctrine'][$mappingType][$serviceType][$serviceName]
)
```
?

greg0ire: The last line looks funny BTW… maybe it should be `$config['doctrine_factories'][$mappingType][$serviceType][$serviceName]`

src/DoctrineModule/Validator/UniqueObject.php
```diff
@@ -1,49 +1,51 @@
...
+    public const ERROR_OBJECT_NOT_UNIQUE = 'objectNotUnique';
 
-    /**
-     * @var array Message templates
-     */
-    protected $messageTemplates = [
-        self::ERROR_OBJECT_NOT_UNIQUE => "There is already another object matching '%value%'",
-    ];
+    /** @var mixed[] */
+    // phpcs:disable
```
greg0ire: What's wrong here?

tests/DoctrineModuleTest/Validator/UniqueObjectTest.php
```diff
@@ -312,6 +312,10 @@ public function testCanFetchIdentifierFromObjectContext()
         $context     = new stdClass();
         $context->id = 'identifier';
 
+#        $context = [
+#            'id' => 'identifier'
+#        ];
+
```
greg0ire: What is this about?

src/DoctrineModule/ServiceFactory/AbstractDoctrineServiceFactory.php
SenseException: Do we need `$config['doctrine_factories'][$mappingType]` when there is `$config['doctrine_factories'][$mappingType][$serviceType]`? Is `isset` still have this behaviour where it can return true despite an in-between element doesn't exist?

greg0ire: Seems useless indeed: https://3v4l.org/DFqf7

src/DoctrineModule/ConfigProvider.php
TomHAnderson: phpcs is disabled here because it violates the >120 char rule.  I could not find that rule as it is not in the Doctrine coding standard ruleset and when I  used `SlevomatCodingStandard.Sniffs.Files.LineLengthSniff` it didn't work.  There are two places where phpcs is disabled globally because of this line length issue.

src/DoctrineModule/Validator/UniqueObject.php
TomHAnderson: I swapped these so the `/** @var mixed[] */` is directly over the `$messageTemplates`.

phpcs is disabled here because it violates the >120 char rule. I could not find that rule as it is not in the Doctrine coding standard ruleset and when I used SlevomatCodingStandard.Sniffs.Files.LineLengthSniff it didn't work.

tests/DoctrineModuleTest/Validator/UniqueObjectTest.php
TomHAnderson: The type hint for $context was an array but the code was passing an object.  These comments were left over from my debugging.  I fixed the problem by removing the type hint.

src/DoctrineModule/ConfigProvider.php
greg0ire: But… the lines were fine (lower than 120 char) before the changes… were the arrays change from multiline to oneline because of some other rule?

TomHAnderson: I've fixed all phpcs to use the proper sniff: https://github.com/doctrine/DoctrineModule/pull/683/commits/f20bd42e563b469d01efb5ff7a90e36143f3e789

src/DoctrineModule/Authentication/Adapter/ObjectRepository.php
TomHAnderson: I just got moved; see line 137

TomHAnderson: Oh, this was a missing `.` at the end of the comment after the RuntimeException.  I'll fix that.

TomHAnderson:
The arrays were changed from multiline to one line because of
`Squiz.Arrays.ArrayDeclaration`
which does not exclude `Squiz.Arrays.ArrayDeclaration.MultiLineNotAllowed`
which does not allow for single array elements to exist on multiple lines.

Do you think this is an oversight in the Doctrine coding-standard?

greg0ire:
> Do you think this is an oversight in the Doctrine coding-standard?

Maybe yeah… this is kind of an edge case

src/DoctrineModule/Options/Cache.php
```diff
@@ -42,78 +42,50 @@ class Cache extends AbstractOptions
...
     }
 
-    /**
-     * @return string
-     */
-    public function getInstance()
+    public function getInstance() : string
     {
-        return $this->instance;
+        return (string) $this->instance;
```
gszy: No no no no no! Please roll this back! :scream:

`$this->instance` can be `null`, but must never be an empty string, see [`DoctrineModule\Service\CacheFactory`, line 40](https://github.com/doctrine/DoctrineModule/blob/4.0.0/src/DoctrineModule/Service/CacheFactory.php#L40).

TomHAnderson: Will this suffice?  https://github.com/doctrine/DoctrineModule/pull/687

gszy: Yes, it would fix it, but please have a look at #686 (I’ve added another typehint in line 41 and kept native PHP typehint in line 64).

PR #690, opened 2020-03-24
Fix a typo in .travis.yml (CHECK_CS instead of CS_CHECK)

gszy:


greg0ire:
And as expected, the build fails because of all the CS issues :P

PR #691, opened 2020-03-24
Update copyright years

gszy:


LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2006-2015 Doctrine Project
+Copyright (c) 2006-2020 Doctrine Project
```
greg0ire: I think the end year can be removed entirely
```suggestion
Copyright (c) 2006 Doctrine Project
```

TomHAnderson: I prefer the range  because is shows attention since the inception.  

It is a pain to update each year and project will go a few years without an update to the end/current year but I'm ok with that.

greg0ire: As you wish :)

greg0ire:
Thanks @gszy !

PR #702, opened 2020-06-29
Authentication: Allow specification of session container/member

alichry:
Hello,

This PR allows specifying the session container name and member to use in the underlying Session Storage instead of using the default. In case one of the new options are set, the StorageFactory creates a local instance of the Session Storage. Otherwise, it will retrieve the default aliased Session Storage `'DoctrineModule\Authentication\Storage\Session'`

The Session Storage instance wrapper (ObjectRepository) is still cached by service manager.


TomHAnderson:
Moved to https://github.com/doctrine/DoctrineModule/pull/712 to change base from 4.1.x to 4.0.x

PR #712, opened 2020-11-13
Custom session storage parameters

TomHAnderson:
This is a cherry-pick of https://github.com/doctrine/DoctrineModule/pull/702 to 4.0.x

### DIFF
--- a/src/DoctrineModule/Service/Authentication/StorageFactory.php
+++ b/src/DoctrineModule/Service/Authentication/StorageFactory.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace DoctrineModule\Service\Authentication;
 
 use DoctrineModule\Authentication\Storage\ObjectRepository;
+use DoctrineModule\Options\Authentication as AuthenticationOptions;
 use DoctrineModule\Service\AbstractFactory;
 use Interop\Container\ContainerInterface;
 use Laminas\Authentication\Storage\Session;
@@ -62,6 +63,6 @@ class StorageFactory extends AbstractFactory
 
     public function getOptionsClass(): string
     {
-        return 'DoctrineModule\Options\Authentication';
+        return AuthenticationOptions::class;
     }
 }
--- a/src/DoctrineModule/Service/Authentication/StorageFactory.php
+++ b/src/DoctrineModule/Service/Authentication/StorageFactory.php
@@ -38,8 +38,8 @@ class StorageFactory extends AbstractFactory
             $sessionMember    = $options->getSessionMember();
             if ($sessionContainer || $sessionMember) {
                 $storage = new Session(
-                    $options->getSessionContainer(),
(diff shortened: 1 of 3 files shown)


TomHAnderson:
From @alichry 

This PR allows specifying the session container name and member to use in the underlying Session Storage instead of using the default. In case one of the new options are set, the StorageFactory creates a local instance of the Session Storage. Otherwise, it will retrieve the default aliased Session Storage 'DoctrineModule\Authentication\Storage\Session'

The Session Storage instance wrapper (ObjectRepository) is still cached by service manager.

TomHAnderson:
Beyond what @alichry submitted I created a unit test for their work.

tests/DoctrineModuleTest/Service/Authentication/StorageFactoryTest.php
```diff
@@ -67,8 +70,46 @@ public function testCanInstantiateStorageFromServiceLocator(): void
...
+                        'sessionContainer' => 'customContainer',
+                        'sessionMember' => 'customMember',
+                        'storage' => 'custom_storage',
+                    ],
+                ],
+            ],
+        ];
+
+        $serviceManager
+            ->expects($this->at(0))
```
greg0ire: `at` is deprecated in recent versions of PHPUnit, you should avoid it or will mean more work when migrating.

If you mean to check that the service manager is called, use

```suggestion
            ->expects($this->once())
```

If you mean to just set up the service manager, then this is a stub, and you can drop the line entirely (and use `createStub` over `createMock`.

Another solution could be to use a real service manager instead of a mock.

TomHAnderson: It is a mock, not a stub.  Changed to use `onConsecutiveCalls` and removed `->at()` calls.

greg0ire (approved):

TomHAnderson:
Rebase to correct commit message for phpcs

src/DoctrineModule/Service/Authentication/StorageFactory.php
```diff
@@ -32,7 +34,18 @@ public function __invoke(ContainerInterface $container, $requestedName, ?array $
 
         $storage = $options->getStorage();
         if (is_string($storage)) {
-            $options->setStorage($container->get($storage));
+            $sessionContainer = $options->getSessionContainer();
+            $sessionMember    = $options->getSessionMember();
+            if ($sessionContainer || $sessionMember) {
+                $storage = new Session(
+                    $options->getSessionContainer(),
+                    $options->getSessionMember()
```
SenseException: The variables should be used here.

SenseException (changes requested):

alichry:
Thank you for adding the required test case and considering this feature in the next release :)

SenseException (approved):

PR #718, opened 2020-11-16
Added .doctrine-project.json

TomHAnderson:


.doctrine-project.json
```diff
@@ -0,0 +1,48 @@
...
+      ]
+    },
+    {
+      "name": "4.1.x",
+      "branchName": "4.1.x",
+      "slug": "4.1.x",
+      "current": false,
+      "maintained": true,
+      "aliases": [
+        "future"
```
SenseException:
```suggestion
        "latest"
```
would be consistent with the rest of the projects.

SenseException (changes requested):

.doctrine-project.json
TomHAnderson: But 4.1.x is not the latest; there have been releases of 4.0.x since.  I would be comfortable adding "latest" to 4.0.x and not using any aliases for 4.1.x  Would that be ok?

TomHAnderson: 4.0.x is stable and 4.1.x is not so using latest seems ok for 4.1.x

SenseException (approved):

greg0ire (approved):

PR #726, opened 2021-06-24
Allow laminas form v3

snapshotpl:


SenseException (commented):
The changes of the return types for the `getValueOptions()` would be a BC break for 4.1. I'd expect these will be fine for the next major version or maybe minor, depending on how BC is handled for the module.

TomHAnderson:
BC breaks are handled with a major version change.  I'll review the current branches and get this merged.

TomHAnderson:
For support of PHP 8 we'll need to upgrade to laminas-cli from laminas-mvc-console / laminas-console.  I'll get started on that work but this PR will be behind that update.

driehle:
@snapshotpl @SenseException @TomHAnderson Please have a look at #731 for a suggestion on how we could support laminas-form v3 without breaking backward compatibility. I would prefer that solution, because it will not require a new major version of DoctrineModule and it will leave it to the users when to migrate to laminas-form v3.

PR #730, opened 2021-09-21
fix #728 : DriverFactory incompatible with AttributeDriver

### LINKED ISSUE #728: DriverFactory incompatible with AttributeDriver
The `DriverFactory` is incompatible with the `AttributeDriver` introduced to support the native attributes in PHP 8. This is because the `AttributeDriver` extends the `AnnotationDriver` and as a result the wrong constructor is assumed.

The error:
> Doctrine\ORM\Mapping\Driver\AttributeDriver::__construct(): Argument #1 ($paths) must be of type array, Doctrine\Common\Annotations\CachedReader given, called in vendor/doctrine/doctrine-module/src/DoctrineModule/Service/DriverFactory.php on line 86 and defined in vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php on line 36

The problematic code (specifically the second part of the `if`-statement):
https://github.com/doctrine/DoctrineModule/blob/b0a1f82690b69e0c075cf6db4527a25d5a6ec633/src/DoctrineModule/Service/DriverFactory.php#L77-L89

fezfez:


TomHAnderson:
Hi @fezfez ,

Your class name is wrong.  Thanks for the PR.  Can you add a unit test too?

fezfez:
hi @TomHAnderson

After checking everywhere, the class does not exist in doctrine/persistance, so i've added doctrine/orm in dev deps to add a test

fezfez:
@TomHAnderson : do you have time to check this and make release soon ?

tests/DoctrineModuleTest/Service/DriverFactoryTest.php
```diff
@@ -61,4 +61,23 @@ public function testCreateDriverChain(): void
         $this->assertArrayHasKey('Foo\Bar', $drivers);
         $this->assertInstanceOf('DoctrineModuleTest\Service\Mock\MetadataDriverMock', $drivers['Foo\Bar']);
     }
+
+    public function testCreateAttributDriver(): void
```
TomHAnderson, line 65: Class name is incorrect

TomHAnderson (changes requested):
Rebase these two commits into a single commit please.

tests/DoctrineModuleTest/Service/DriverFactoryTest.php
driehle:
```suggestion
    public function testCreateAttributeDriver(): void
```

driehle:
@fezfez Could you please rebase into a single commit and fix the class name?

driehle:
Closing in favour of #734 as fezfez has disabled edits from maintainers in this PR.

PR #731, opened 2021-10-06
Allow laminas form 3.0 while keeping BC

driehle:
This PR is an alternative to #726. In contrast to #726 this PR keeps backward compatibility by using polyfills. Technically, there are now two different implementations of the form elements, one targeted at laminas-form v2 and a different one targeted at laminas-form v3. Using Composers autoload feature, the implementation corresponding to the installed laminas-form version is loaded and put into the right place using `class_alias`.

In case you are skeptical about this technique, I'd like to refer to laminas-form 2.15.x series, which itself used the same principle to support both laminas-servicemanager v2 and v3. Have a look at their [composer.json](https://github.com/laminas/laminas-form/blob/2.15.x/composer.json#L65), [autoload file](https://github.com/laminas/laminas-form/blob/2.15.x/autoload/formElementManagerPolyfill.php) and [polyfills](https://github.com/laminas/laminas-form/tree/2.15.x/src/FormElementManager).


driehle:
Please note that I have updated the Github workflows:
 - Unit tests are now running with both highest and lowest dependencies for all PHP versions, this effectively tests both laminas-form 2.x and 3.x
 - Coverage reports are uploaded to Codecov again (seems like no-one has updated that since Travis shut down...)

snapshotpl (approved):

.github/workflows/coding-standards.yml
```diff
@@ -0,0 +1,38 @@
...
+          tools: "cs2pr"
+
+      - name: "Install dependencies with Composer"
+        uses: "ramsey/composer-install@v1"
+        with:
+          dependency-versions: "highest"
+
+      # https://github.com/doctrine/.github/issues/3
+      - name: "Run PHP_CodeSniffer"
+        run: "vendor/bin/phpcs -q --no-colors --report=checkstyle | cs2pr"
```
greg0ire: Please configure your IDE to always add a newline at end of file, this will make the red pictogram disappear.

.github/workflows/coding-standards.yml
```diff
@@ -0,0 +1,38 @@
+name: "Coding Standards"
```
greg0ire, line 1: Let's reuse the centralized workflow here?

.github/workflows/continuous-integration.yml
```diff
@@ -8,72 +6,70 @@ on:
...
-      - name: "Install dependencies with composer"
-        run: "composer update --no-interaction --no-progress --no-suggest"
+          path: "reports"
 
-      - name: "Run PHPUnit"
-        run: "composer test-coverage"
+      - name: "Upload to Codecov"
+        uses: "codecov/codecov-action@v1"
+        with:
+          directory: "reports"
```
greg0ire, line 82: This looks generic enough that you could reuse the centralized workflow here too.

.github/workflows/coding-standards.yml
driehle: Sorry, haven't configured all workstations accordingly yet. Will be fixed.

.github/workflows/coding-standards.yml
driehle: Agree

.github/workflows/continuous-integration.yml
driehle: I'll want to add different testing for laminas-form ^2.0 and ^3.0, similar to what is in `doctrine/orm` as you suggested. I think then I can't use the centralized workflow anymore?

src/DoctrineModule/Form/Element/ObjectMultiCheckboxV3Polyfill.php
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * @param mixed $options
+     */
+    public function setOptions($options): self
+    {
+        $this->getProxy()->setOptions($options);
+
+        return parent::setOptions($options);
+    }
```
SenseException: Does it make sense to move identical methods into a trait too?

driehle: I think you can't call `parent` from within a trait, can you? I remember I tried implementing it that way and there was this issue.

src/DoctrineModule/Form/Element/HasProxy.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineModule\Form\Element;
+
+trait HasProxy
```
TomHAnderson: I dislike the name of this trait.  HasProxy implies a boolean but `getProxy` will always return a Proxy.  I would prefer this be renamed to `trait GetProxy`.

src/DoctrineModule/Form/Element/ObjectMultiCheckboxV3Polyfill.php
```diff
@@ -0,0 +1,72 @@
...
+use Laminas\Form\Element\MultiCheckbox;
+use Laminas\Stdlib\ArrayUtils;
+use Traversable;
+
+use function array_map;
+use function is_array;
+
+class ObjectMultiCheckboxV3Polyfill extends MultiCheckbox
+{
+    use HasProxy;
```
TomHAnderson, line 16: Change to `GetProxy`

TomHAnderson (changes requested):

src/DoctrineModule/Form/Element/HasProxy.php
TomHAnderson: Apply change to all polyfills

driehle: Trait now renamed to `GetProxy`

src/DoctrineModule/Form/Element/ObjectMultiCheckboxV3Polyfill.php
driehle: And there are discussions, that a trait should never override a method, see e.g. https://softwareengineering.stackexchange.com/questions/371067/should-a-trait-refer-to-parent-methods. Therefore, I would like to refrain from moving this into a trait somehow.

TomHAnderson (approved):

driehle:
Rebased on current `4.2.x`. @SenseException are you fine with the trait?

PR #732, opened 2021-10-12
Fix configuration for laminas-component-installer

driehle:
The `composer.json` currently still holds an extra configuration targeted at the Zend Framework component installer, which was not ported to Laminas correctly. This PR holds the changes for porting this extra configuration to Laminas.

Note that in contrast to DoctrineModule (this repository), DoctrineORMModule was updated correctly: https://github.com/doctrine/DoctrineORMModule/blob/4.0.x/composer.json#L44

Besides, the branch aliases for 1.2-dev and 2.0-dev are outdated and should removed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,13 +36,9 @@
         }
     ],
     "extra": {
-        "zf": {
+        "laminas": {
             "config-provider": "DoctrineModule\\ConfigProvider",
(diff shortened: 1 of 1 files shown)


greg0ire:
Thanks @driehle !

PR #737, opened 2021-10-18
Switch to PSR-4 and move classes accordingly

driehle:
Since [doctrine-laminas-hydrator](https://github.com/doctrine/doctrine-laminas-hydrator) is and has always been using PSR-4, this project should follow. This PR switches from PSR-0 to PSR-4 and moves all classes in the right places, effectively droping one level of folder hierarchy in both `src` and `test`. 

I'd like to target this at the next minor release, i.e. 4.2.0.


composer.json
```diff
@@ -78,13 +78,13 @@
...
-        "psr-0": {
+        "psr-4": {
             "DoctrineModule\\": "src/"
         }
     },
     "autoload-dev": {
-        "psr-0": {
-            "DoctrineModuleTest\\": "tests/"
+        "psr-4": {
+            "DoctrineModuleTest\\": "test/"
```
greg0ire: This change is not justified or mentioned in the description of your PR / in your commit message.
I don't see why we should diverge from other Doctrine repositories on this (or from https://github.com/php-pds/skeleton).
Also, the change is incomplete, there is an occurrence of `tests` here:
https://github.com/doctrine/DoctrineModule/blob/662362af298d25a305d62e96931c0a3608078b40/.gitattributes#L1

greg0ire (commented):
Good move 👍

composer.json
driehle: In `doctrine-laminas-hydrator` and `DoctrineORMModule` the folder name is `test`. I wanted to keep it the same. In `DoctrineMongoODMModule` it is `tests`, we can alternatively change everything to `tests`. Would you prefer that?

driehle: As discussed on Slack, we'll keep `tests` here and adapt the other repositories accordingly.

.gitignore
```diff
@@ -1,8 +1,10 @@
...
-vendor/
-phpunit.xml
-.project/
-nbproject/
-composer.lock
-.DS_Store
-.phpunit.result.cache
+/.buildpath
+/.project
+/.idea
```
greg0ire: Please, let's remove OS and IDE-specific directories, the list can grow indefinitely otherwise. People should use a global gitignore for this instead: https://greg0ire.fr/git-gud/#/23

TomHAnderson:
@Ocramius I wanted to do this years ago but was told then that the Doctrine standard is PSR-0.  Has that changed?  Can we go ahead with PSR-4 (this has my recommendation)

greg0ire:
We've had a discussion about this BTW: https://github.com/orgs/doctrine/teams/doctrinecore/discussions/27

driehle:
Plus, doctrine-laminas-hydrator has been using PSR-4 ever since and I strongly dislike having different standards across the Laminas-related projects. Porting doctrine-laminas-hydrator from PSR-4 to PSR-0, something which is essentially dead, does not make any sense to me.

.gitignore
driehle: Updated

.gitignore
```diff
@@ -1,8 +1,7 @@
...
-.project/
-nbproject/
-composer.lock
-.DS_Store
-.phpunit.result.cache
+/.phpcs.cache
+/.phpunit.result.cache
+/clover.xml
+/composer.lock
+/phpcs.xml.dist
```
TomHAnderson: I don't think `phpcs.xml.dist` should be in the ignore list.

TomHAnderson (changes requested):

.gitignore
driehle: Indeed, there should be `phpcs.xml` instead of `phpcs.xml.dist`, ~~and there should be `phpunit.xml` as well, which is currently still missing. I'm going to fix this.~~ All good, `phpunit.xml` is there. Only the issue with `phpcs.xml`.

TomHAnderson: I disagree.  `phpcs.xml` and `phpunit.xml` are the purview of the developer and not the package.  It is correct to ignore them in .gitignore and distribute the `.dist` versions.

TomHAnderson (approved):

.gitignore
driehle: Yes, excactly what I meant, but they way I wrote it down was misleading, sorry. I meant that there should be `phpcs.xml` and `phpunit.xml` on the `.gitignore`, but it was easy to understand the other way round. :-)

PR #738, opened 2021-10-21
Update dependencies and switch to PHPUnit 9

driehle:
For the next minor release (i.e. 4.2.0) I'd like to update all dependencies to their latest version that is still supporting PHP 7.3. Except for `laminas-servicemanager`, all dependencies still support PHP 7.3. The service manager has to stay at 3.7.0, as 3.8.0 dropped PHP 7.3 there.

With the update of the dependencies, I switched from PHPUnit 8 to PHPUnit 9. This brings in some deprecation messages, mostly about `prophesize()`. The package `phpspec/prophecy-phpunit` was added to solve this. Another issue refered to `assertRegExp()` (now named `assertMatchesRegularExpression()`) and `expects($this->at(1)` not being supported anymore (solved by using `willReturnMap()` instead).

The `phpcs.xml.dist` was updated to be in line with the one a `doctrine/doctrine-laminas-hydrator`. Actually, there were a couple of old exceptions in that file, which - when removed - did not cause any issues. Hence it makes sense removing them.


phpcs.xml.dist
```diff
@@ -1,42 +1,21 @@
...
-        <exclude-pattern>*</exclude-pattern>
-    </rule>
-
-    <!-- Paths to check -->
+    <!-- Directories to be checked -->
     <file>src</file>
     <file>tests</file>
+
+    <!-- Include full Doctrine Coding Standard -->
+    <rule ref="Doctrine"/>
```
snapshotpl, line 20: How phpcs know where to find this standard and where is it come from?

greg0ire: Good questions!

- https://github.com/doctrine/coding-standard/blob/f86c16aedb8a237f6c3deff01bd0131ac08d4570/composer.json#L25
- https://github.com/Dealerdirect/phpcodesniffer-composer-installer#how-it-works

snapshotpl: So magic, but works ;-)

snapshotpl (approved):

greg0ire (approved):

SenseException (approved):

PR #739, opened 2021-10-21
Enable unit tests to work with PHP 8.0

### LINKED ISSUE #740: Unit tests should be executed with PHP 8.0
I think I might have found a suitable approach for getting unit tests running with PHP 8.0. 

I'd like to discuss the changes proposed in https://github.com/doctrine/DoctrineModule/pull/739 with you. First, as discussed on Slack, we are using symbols from laminas-(mvc-)console in `src` without required that package explicitly (it is in `suggest`, though). Therefore, at first, I have identified these placed, added an appropriate check on top of the file and use `trigger_error` to generate a runtime error in case someone tries using that classes without having the suggestion installed.

Next, I have deprecated the related symbols so that we can remove them with the next major version. To allow running the unit tests with PHP 8, I used `--ignore-platform-req=php`. I know that this is not the best choice, but I think we don't really have a chance to do otherwise. Besides laminas-(mvc-)console, there is laminas-cache as well, which still isn't fully PHP8-compatible (at lest withouth manually replacing legacy cache adapters, see docs).

Personally, I think we can go this way. I cannot think of a scenario where this will break someones application. We are continuing to support under PHP 7.3 & 7.4 what we have supported before and we are continuing not to support under PHP 8.0 what we have never supported before. What do you guys think? 

cc @TomHAnderson @greg0ire

driehle:
This PR improves PHP 8.0 compatibility. `DoctrineModule` was already compatible with PHP 8.0 before, but there were some edge cases with the optional dependency `laminas/laminas-mvc-console`, which have not been handled.

With this PR, explicit checks for the presence of `laminas/laminas-mvc-console` have been added and errors are reported using `trigger_error()`, when non-installed components are being used.

Further, this PR removed the method `\DoctrineModule\Module::getConsoleUsage()` when PHP 8.0 is used. This method is used by `laminas/laminas-mvc-console`, which is not PHP 8.0-compatible. Hence, this method can never have been used when using PHP 8.0. For PHP 7.3 and 7.4 everything will continoue to work as before. We have discussed this issue on Slack in in the comments here and came to the conclusion that we do not consider this a BC break.

This PR targets the next minor release, i.e. 4.2.0.

Fixes #740


src/Module.php
```diff
@@ -68,8 +71,17 @@ public function getConfig()
     /**
      * {@inheritDoc}
      */
-    public function getConsoleUsage(Console $console)
+    public function getConsoleUsage($console)
```
greg0ire: Is this class likely to be extended? Because this is a BC-break for extending classes.

SenseException: The lowest version is PHP 7.3, so it should be possible to omit the type hint in the class that extends `Module`. The `Console` type hint can be kept in there. Why was it removed anyway?

greg0ire: @SenseException I think you are getting it backwards: https://3v4l.org/CACMI

> Why was it removed anyway?

Good question though :)

snapshotpl: Maybe because Laminas\Console is now deprecated?

SenseException: @greg0ire I meant it that way: https://3v4l.org/o0TVq

greg0ire: @SenseException yeah, that's why I suggested you got it backwards: in this case was talking about extending classes, not extended classes (or implemented interfaces). Doing this breaks the LSP, that's why even PHP 8 does not allow it.

driehle: @greg0ire @SenseException I removed it, because this class doesn't exist in PHP 8 (the package providing it is not PHP 8 compatible). So if I keep it there I think it will throw an error in PHP 8.

If you think this is a BC break, and from what you say this could actually be true, then I could use two different polyfills for this class. What do you think?

driehle: Or, alternatively, maybe more simple, define `Laminas\Console\Adapter\AdapterInterface` (i.e. Console) as an empty interface if and only if PHP 8 and it doesn't exist?

greg0ire: I'm confused: if the class does not exist in PHP 8, then what arguments is calling code going to pass to this method?

driehle: Nothing. That method is never gonna be called using PHP 8. Laminas-console calls this message to render a help information / usage information (like `something -h`) to the console output. Since laminas-console (https://packagist.org/packages/laminas/laminas-console) is `^5.6 || ^7.0` you won't be able to run that with PHP 8, unless you do some hacking.

However, obviously someone could call this method manually for whatever reason. If that person was using PHP 7, everything is fine, as it will continue to work. If that person was using PHP 8, then the method will return an empty string. Still, in the latter case, the person must have done some type of hacking (for instance, `--ignore-platform-arg=php`), otherwise you wouldn't be able to install laminas-console under PHP 8.

(Side note: laminas-mvc-console, the package we are suggesting/have in require-dev, see https://packagist.org/packages/laminas/laminas-mvc-console, is actually PHP 8 compatible, but it requires laminas-console right away, so in the end, is is still not PHP 8 compatible!)

greg0ire: > However, obviously someone could call this method manually for whatever reason.

This sounds unlikely. Maybe put this method in a trait that is empty for PHP 8? Example: https://github.com/doctrine/dbal/blob/2.13.x/lib/Doctrine/DBAL/Driver/PDOQueryImplementation.php

driehle: Indeed it is very unlikely. It is rather a question of what we consider a BC break and what not. I like your idea of using a trait, will do that!

src/Controller/CliController.php
```diff
@@ -10,10 +10,25 @@
 use Symfony\Component\Console\Application;
 use Symfony\Component\Console\Output\OutputInterface;
 
+use function class_exists;
 use function is_numeric;
+use function sprintf;
+use function trigger_error;
+
+if (! class_exists(AbstractActionController::class)) {
```
snapshotpl: This is `Laminas\Mvc\Controller\AbstractActionController` not from `Laminas\Mvc\Console` namespace

src/GetConsoleUsage.php
```diff
@@ -0,0 +1,57 @@
...
+     *                   Please use ./vendor/bin/doctrine-module instead.
+     */
+    trait GetConsoleUsage
+    {
+    }
+} else {
+    /**
+     * @internal
+     * @deprecated 4.2.0 Usage of laminas/laminas-mvc-console is deprecated, integration will be removed in 5.0.0.
+     *                   Please use ./vendor/bin/doctrine-module instead.
```
snapshotpl, line 31: I'm not sure is good replacement. I think we need to use `laminas-cli` instead

src/Validator/Service/AbstractValidatorFactory.php
```diff
@@ -98,10 +98,10 @@ protected function merge(array $previousOptions, array $newOptions): array
     }
 
     /**
-     * Helper method for ZF2 compatiblity.
+     * Helper method for Laminas compatiblity.
      *
-     * In ZF2 the plugin manager instance if passed to `createService`
-     * instead of the global service manager instance (as in ZF3).
+     * In Laminas v2 the plugin manager instance if passed to `createService`
+     * instead of the global service manager instance (as in Laminas v3).
```
snapshotpl, line 103: If service-manager v3 is now minimal supported version, then we don't need this helper (and also `createService` method)

src/GetConsoleUsage.php
driehle: In the long run it would make sense to use `laminas-cli`, but then users would need to run `./vendor/bin/laminas orm:schema-tool:create` instead of `./vendor/bin/doctrine-module orm:schema-tool:create`, which is not much of a difference. We do currently already have our own CLI (`./vendor/bin/doctrine-module`), so I think there is no need for a rush implementing `laminas-cli`. Technically, `laminas-cli` builds on top of `symfony/console` and so does our own `doctrine-module` bin script. So in the end, they would really be the same.

src/Controller/CliController.php
driehle: Ouch, thanks. Good catch.

driehle: Actually this is good, then the check can go into `cliAction()`, as the symbols from `laminas-mvc-console` are only used within that method, but nowhere in signatures.

src/Validator/Service/AbstractValidatorFactory.php
driehle: Indeed, but as `AbstractValidatorFactory` is not marked `internal`, I think we cannot drop it in a minor release. Same applies for the `createService()` method. We should, however, mark it as deprecated. What do you think?

src/Controller/CliController.php
driehle: Fixed in 4227418

src/Validator/Service/AbstractValidatorFactory.php
driehle: See 4227418 for what I think we can do in a minor release. Next major will drop all deprecated functions without replacement. Background: Someone may have extended `AbstractValidatorFactory` and can rely on the presence of e.g. `createService()`.

src/GetConsoleUsage.php
snapshotpl: Make sens for me

src/Validator/Service/AbstractValidatorFactory.php
snapshotpl: Like this solution.

snapshotpl (approved):

src/Controller/CliController.php
```diff
@@ -40,6 +46,13 @@ public function setOutput(OutputInterface $output): void
      */
     public function cliAction()
     {
+        if (! class_exists(AbstractActionController::class)) {
```
snapshotpl: I think it's still wrong here. You check now `laminas/laminas-mvc` not `laminas/laminas-mvc-console` as mentioned below.

driehle: Fixed with 9738406

snapshotpl: Perfect!

snapshotpl (approved):

TomHAnderson (approved):

SenseException (approved):

driehle:
Rebased before merging

PR #741, opened 2021-10-22
Declare compatibility with PHP 8.1

driehle:
This PR declares the module to be compatible with PHP 8.1 and enables unit test to run with PHP 8.1, both with laminas-form v2 and v3.

Please note: Some of our upstream dependencies (e.g. laminas-authentication and laminas-cache) are not yet compatible with PHP 8.1. However, this is kind of an chicken vs. egg problem. I suggest that, once our code sucessfully runs with PHP 8.1, we should declare 8.1-compatibility. This is the case now.

In Github actions the dependencies are installed with `--ignore-platform-arg=php`, which tells Composer to ignore (only) the PHP-compatibility declared by our upstream dependencies. This lets us test our module against PHP 8.1

End users will not yet be able to install the module with PHP 8.1 though.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -37,6 +37,14 @@ jobs:
             composer-options: "--ignore-platform-req=php"
             dependencies: "highest"
             laminas-form-version: "3.0.0"
+          - php-version: "8.1"
+            composer-options: "--ignore-platform-req=php"
+            dependencies: "highest"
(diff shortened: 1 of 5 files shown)


greg0ire (approved):

snapshotpl (approved):

SenseException (approved):

driehle:
Will need to merge #739 first, then rebase and merge this one.

PR #748, opened 2021-11-04
Fixed broken require_once in Module.php

driehle:
This issue comes from the switch of PSR-0 to PSR-4, where this file has not been updated accordingly.

The file `Module.php` in the root directory is only used in some cases by laminas/laminas-loader when autoloading via composer is not in place. Nevertheless, this should be fixed anyways.


snapshotpl (approved):

greg0ire (approved):

greg0ire:
Thanks @driehle !

PR #751, opened 2021-11-10
Add missing dependencies to composer.json to prevent unwanted soft dependencies

driehle:
Using [composer-require-checker](https://github.com/maglnet/ComposerRequireChecker) I checked for symbols of libraries which are not in our `require` section of `composer.json`. Such dependencies are called soft dependencies and are usually a bad thing, because their version might change because of changes in upstream libraries, hence, causing bugs for end users. 

To prevent such scenarios, this PR turns the following soft dependencies into hard dependencies:
- container-interop/container-interop
- laminas/laminas-eventmanager
- laminas/laminas-servicemanager

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,12 +52,15 @@
         "doctrine/persistence": "^2.2.2",
         "laminas/laminas-authentication": "^2.8.0",
         "laminas/laminas-cache": "^2.13.0",
+        "laminas/laminas-eventmanager": "^3.4.0",
         "laminas/laminas-form": "^2.17.0 || ^3.0.1",
+        "laminas/laminas-modulemanager": "^2.11.0",
         "laminas/laminas-mvc": "^3.3.0",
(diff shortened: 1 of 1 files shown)


SenseException:
There seem to be a few classes/interfaces in use from `laminas/laminas-console` (e.g. in `DoctrineModule\Component\Console\Input\RequestInput`. Its readme file labels this package as deprecated and was archived though. Isn't this still a soft dependency?

driehle:
You a right that symbols from `laminas/laminas-(mvc-)console` are used, but that is an optional dependency of DoctrineModule. That dependency is listed in the `suggests` section of `composer.json`. Whenever a library uses symbols from an optional dependency, it explicitly needs to check wheter that depenency is installed or not. So a soft dependency is a bad thing if it is a soft dependency by accident (which this PR fixes), but a soft dependency is valid if that is an optional dependency noted in `suggests`.

greg0ire (approved):

PR #753, opened 2021-11-15
Support for Laminas Cache v3

snapshotpl:


.github/workflows/continuous-integration.yml
```diff
@@ -53,6 +59,13 @@ jobs:
         if: "startsWith(matrix.php-version, '8')"
         run: "composer require laminas/laminas-form ^${{ matrix.laminas-form-version }} --no-update --ignore-platform-req=php"
 
+      - name: "Require specific laminas-cache version"
+        run: "composer require laminas/laminas-cache ^${{ matrix.laminas-cache-version }} --no-update"
+
+      - name: "Require specific laminas-cache-storage-adapter-blackhole version"
+        if: "matrix.laminas-cache-blackhole-version"
+        run: "composer require laminas/laminas-cache-storage-adapter-blackhole ^${{ matrix.laminas-cache-blackhole-version }} --no-update"
```
driehle: Why is this step and the parameter `laminas-cache-blackhole-version` only needed in the PHP=7.3, dependencies=lowest setting and not in any of the other settings?

composer.json
```diff
@@ -67,6 +67,8 @@
         "doctrine/coding-standard": "^9.0.0",
         "doctrine/orm": "^2.10.2",
         "jangregor/phpstan-prophecy": "^1.0.0",
+        "laminas/laminas-cache-storage-adapter-blackhole": "^1.2.1 || ^2.0.0",
+        "laminas/laminas-cache-storage-adapter-memory": "^1.0.1 || ^2.0",
```
driehle:
```suggestion
        "laminas/laminas-cache-storage-adapter-memory": "^1.0.1 || ^2.0.0",
```

.github/workflows/continuous-integration.yml
snapshotpl: Require `laminas-cache` in v3 and lowest dependencies update this adapter to v1 which missing config provider, so it's not possible to correctly register in service-manager out of the box.

snapshotpl: @boesing Is it a feature or bug in `laminas-cache`? How to correctly handle tests v2 and v3?

boesing: I have birthday today, I'll have a look in the next days.

I would create some kind of "prepare servicemanager" thing which will either use the delegator from the Adapter to delegate the pluginmanager or not, if not available. 
then using the new storage factory in both v2 and v3.
how to use the delegator directly can be found here:
https://github.com/laminas/laminas-cache-storage-deprecated-factory/blob/e82a23a09f66c7c4a4b234bff418b1624a1c1bf1/src/StorageFactory.php#L265

so in the end your service manager will prepared and you can use it the same way in v2 and v3. Only check you have to make is "is delegator available and if so, add it to the service manager". No need to create a complex matrix.

driehle: @boesing Happy Birthday! 🎂 And thanks for your great work in laminas-cache!

We can get back to this at another day. Maybe it could be as simple as adding a conflict to the cache adapters <2 in the v3 package?

boesing: Just to understand this correct:
When exactly do we need a conflict? 

Require `laminas-cache` as ^2.13.2 and ^3.0 and blackhole as v1 and v2 should do the magic already, not?

both packages need to be explicitly required in case you want to add support for v2 and v3 

if there is a problem due to the lack of `require laminas-cache` in v1, please create a PR in the storage adapter with a conflict so we can merge this. I think we have to add >3.0 here, right? https://github.com/laminas/laminas-cache-storage-adapter-blackhole/blob/4af1053efd81785a292c2a9442871c075700345a/composer.json#L13

driehle: The following composer command runs without any errors and I am not sure if this is how it should be:

```bash
$ composer require laminas/laminas-cache:^3.0 laminas/laminas-cache-storage-adapter-blackhole:^1.0
  - Locking container-interop/container-interop (1.2.0)
  - Locking laminas/laminas-cache (3.0.0)
  - Locking laminas/laminas-cache-storage-adapter-blackhole (1.2.1)
  - Locking laminas/laminas-eventmanager (3.4.0)
  - Locking laminas/laminas-servicemanager (3.10.0)
  - Locking laminas/laminas-stdlib (3.6.1)
  - Locking psr/cache (1.0.1)
  - Locking psr/container (1.1.2)
  - Locking psr/simple-cache (1.0.1)
  - Locking symfony/polyfill-ctype (v1.23.0)
  - Locking webmozart/assert (1.10.0)
```

In essence, you can install the storage satellite packages in the 1.x series with the laminas-cache version 3.x. 
- If this is desired, then we should not have the explicit setting of the blackhole adapter 2.x in line 65-67 and should indeed actually test laminas-cache v3 with the blackhole adapter 1.x.
- If this is not desired, then - from my point of view - laminas-cache 3.x should add conflicts to all storage satellite packages <2.0.

snapshotpl: Look here https://github.com/doctrine/DoctrineModule/runs/4227035949?check_suite_focus=true
It's what happen after disable this step.

Happy Birthday! :gift:

boesing: No when using require, that should install v2. 
can you run `composer why-not laminas/laminas-cache-storage-adapter-blackhole 2.0`?

driehle: Sure, but that will likely not be interesting:

```bash
$ composer why-not laminas/laminas-cache-storage-adapter-blackhole 2.0
__root__  -  requires  laminas/laminas-cache-storage-adapter-blackhole (^1.0)
```

Note that I have explicitly installed `laminas-cache-storage-adapter-blackhole:^1.0` before, therefore this result.

____

The reason why this issue here is happening is that @snapshotpl wanted to support both laminas-cache 2.x and 3.x in DoctrineModule. This leaves the decision which version to use to the end user. Hence, the composer.json looks like this:

```json
"laminas/laminas-cache": "^2.13.0 || ^3.0.0",
"laminas/laminas-cache-storage-adapter-blackhole": "^1.2.1 || ^2.0.0",
```

Now we run unit tests for both laminas-cache 2.x and laminas-cache 3.x per Github actions matrix. This is done by updating the composer.json to either require laminas-cache 2.x or 3.x explicitly (hence, the `||` disappears). As a result, the composer.json using within the unit tests for the 3.x series then looks like this:

```json
"laminas/laminas-cache": "^3.0.0",
"laminas/laminas-cache-storage-adapter-blackhole": "^1.2.1 || ^2.0.0",
```

And installing this with `--prefer-lowest` leads to the error we are having here.

boesing: I have to test that at home but having both versions for both packages should be absolutely okay for PHP <8.1 

not sure why `prefer-lowest` would install v3 of laminas-cache when v2.13 & v3 are allowed 🤷🏼‍♂️ but hard to debug via phone 😅

driehle: We want to test both scenarios explicitly in CI, hence, we do not rely on `prefer-lowest` to test laminas-cache. Instead, we are testing:

- laminas-cache ^2.13 with `prefer-dist`
- laminas-cache ^2.13 with `prefer-lowest`
- laminas-cache ^3.0 with `prefer-dist`
- laminas-cache ^3.0 with `prefer-lowest`

However, the requirement of blackhole always stays `"^1.2.1 || ^2.0.0"`. That leads to the following installations:

- laminas-cache ^2.13 with `prefer-dist` -> installs blackhole 1.0 (because `^1.0` is required in laminas-cache)
- laminas-cache ^2.13 with `prefer-lowest` -> installs blackhole 1.0
- laminas-cache ^3.0 with `prefer-dist` -> installs blackhole 2.0
- laminas-cache ^3.0 with `prefer-lowest` -> installs blackhole 1.0 -> ERROR

If you say that a user error, I fine with it. This, however, implies that users need to now which versions of storage adapters are compatible with with specific version of laminas-cache.

Hence, I'd suggest that laminas-cache should add to its composer.json the following conflicts:

```json
    "conflict": {
        "laminas/laminas-cache-storage-adapter-apc": "<2.0",
        "laminas/laminas-cache-storage-adapter-apcu": "<2.0",
        "laminas/laminas-cache-storage-adapter-blackhole": "<2.0",
        "laminas/laminas-cache-storage-adapter-dba": "<2.0",
        "laminas/laminas-cache-storage-adapter-ext-mongodb": "<2.0",
        "laminas/laminas-cache-storage-adapter-filesystem": "<2.0",
        "laminas/laminas-cache-storage-adapter-memcache": "<2.0",
        "laminas/laminas-cache-storage-adapter-memcached": "<2.0",
        "laminas/laminas-cache-storage-adapter-memory": "<2.0",
        "laminas/laminas-cache-storage-adapter-mongodb": "<2.0",
        "laminas/laminas-cache-storage-adapter-redis": "<2.0",
        "laminas/laminas-cache-storage-adapter-session": "<2.0",
        "laminas/laminas-cache-storage-adapter-wincache": "<2.0",
        "laminas/laminas-cache-storage-adapter-xcache": "<2.0",
        "laminas/laminas-cache-storage-adapter-zend-server": "<2.0"
    },
```

Then, installing laminas-cache 3 with adapters not compatible with v3 would not be possible anymore. I can provide a PR if that is desired.

driehle: Oh, I just saw that we can also look at the problem from the other end:

All cache adapters in 2.x have `laminas-cache:^3.0` as a requirement. So there, the cache adapters specify which laminas-cache version they are compatible with.

The adapters in 1.x do NOT have any requirement towards laminas-cache. Maybe that is the actual problem here.

(Still, since all the old 1.x storage packages are already released, I think you can only solve it by adding above mentioned conflict in laminas-cache v3. If you'd update the old 1.x packages, then composer would simply install earlier versions, which do not have that requirement.)

boesing: yes, that was due to circular dependency in v2 where laminas-cache had still require all adapters.
What we can do is to add conflict in all adapter v1 versions with 3.0 of laminas-cache.

driehle: Well, if you add the conflict, for example, in `laminas-cache-storage-adapter-blackhole`, then you would release a version 1.2.2 with conflict laminas-cache >= 3.0. Then, however, composer would simply install 1.2.1, which does not have that said conflict.

I think you can only add the conflict on the side of laminas-cache v3.

boesing: No, if you properly declare your requirements (`^1.2.2 || ^2.0`), you are totally fine and composer won't simply install `1.2.1` as that is not allowed by your definition here.

What I do not understand is, why won't you simply check the three cases (as we do in laminas-* components with our matrix & CI)?

These would be:
- lowest
- locked
- latest

When you have this combination:
```
{
    "require": {
        "laminas/laminas-cache": "^2.13.2 || ^3.0",
        "laminas/laminas-cache-storage-adapter-blackhole": "^1.2.1 || ^2.0"
    }
}
```

This would run (depending on when the tests are being executed) with both versions (and probably newer versions of blackhole & laminas-cache which are not yet released).

That would be way enough for a component.
No need to test "lowest, latest" for both "2.x" and "3.x" of laminas-cache and/or "1.x" and "2.x" for the blackhole adapter.
I've never seen `composer require` is being used in any other component during CI to install specific versions, why would you even do that?
Require laminas-cache along with blackhole (as `require-dev`) and you are good to go.

Adding specific cache adapters to the `require` section would be fully against the idea of laminas-cache v3 where upstream projects should declare which adapters they want to use.

boesing: BTW: the "problem" with laminas-cache v3 is, that if you do not specify an adapter (which would be the right way), this module won't be installable in projects which do not use laminas-cache at all.
Because `laminas-cache` in v3 needs at least one adapter available. You can, ofc, add the `blackhole` or `memory` adapter to this projects `require` so an installable set is allways available.

snapshotpl: It's done because convention like with `laminas-form`. Release `1.2.2` with conflict (or restricted version 2 of `laminas-cache`) should help here.

boesing: you can then simply require both Adapter versions which are compatible with the appropriate cache version. actually, you simply require laminas-cache but what if you require Adapter and cache in your action setup step?

boesing: I had a quick glance of this in a local branch and I also realized config incompatibility. I havent realized that earlier so I will see why this is happening. the StorageFactory in v2 requires `adapter.name` which is not compatible with the config the StorageFactoryInterface is expecting.

I'd say this package needs a little rewrite in the cache factory so it uses the new storage factory rather than depending on the abstract factory. 

I will try to see if I can make that happen in the next few days and present what I have in mind. 👍🏻

snapshotpl:
@boesing I have updated `laminas-cache` to 3.1 but looks like `StorageAdapterFactory.php` doesn't allow `adapter` or I'm doing something wrong

.github/workflows/continuous-integration.yml
snapshotpl: Config for StorageFactory fixed. For me we should stay with this check since not fixed in `laminas-cache`. @driehle WDYT?

driehle (approved):

snapshotpl:
@driehle How can I help with release 4.3?

driehle:
@snapshotpl 4.3.0 is now released :)

PR #754, opened 2021-11-17
laminas cache 3.0

boesing:


tests/Service/CacheFactoryTest.php
```diff
@@ -64,12 +68,18 @@ public function testCreateLaminasCache(): void
                 ],
                 'caches' => [
                     'my-laminas-cache' => [
-                        'adapter' => ['name' => 'blackhole'],
+                        'adapter' => 'memory',
```
driehle, line 71: Wouldn't this technically be a BC break for DoctrineModule users, if the old configuration syntax doesn't work anymore?

boesing: it does on v2, it wont on v3. 
thats not an issue for doctrine Users if they never used laminas-cache. if they do, they should require the cache version themselves. 
if you have different oppinions on this, you might consider releasing a new major instead while dropping support for v2 and referring to the laminas-cache config style (documentation).

boesing:
this PR should only be an idea so #753 can be finished. I would not expect this PR to be merged.

driehle:
Thanks for your input @boesing!

PR #755, opened 2021-11-29
Add missing provide in composer.json

snapshotpl:
Without this section user cannot install laminas-cache alone (this package provide implementation)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -63,6 +63,9 @@
         "laminas/laminas-validator": "^2.15.0",
         "symfony/console": "^5.3.7"
     },
+    "provide": {
+        "laminas/laminas-cache-storage-implementation": "1.0"
+    },
     "require-dev": {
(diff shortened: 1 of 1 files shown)


driehle:
Since we don't have any storage adapter in `require` section of `composer.json` (only in `require-dev`), I don't thnk we should add the suggested `provide` configuration. 

We either need to update docs saying that users need to require their own storage adapters together with DoctrineModule, or we'd need to require at least the memory cache adapter (and probably the filesystem adapter as well). Anyways, docs should probably be updated either way.

snapshotpl:
What about this adapter https://github.com/doctrine/DoctrineModule/blob/4.3.x/src/Cache/DoctrineCacheStorage.php ?

driehle:
Ok, forgot about that adapter 👍

driehle (approved):

driehle:
Thanks @snapshotpl!

PR #759, opened 2021-12-14
Remove --ignore-platform-req=php

snapshotpl:


driehle:
This won't work for 4.x I think. However, I'd support work towards a 5.0.0 release, which drops laminas-(mvc-)-console. Then, we should be able to get rid of `--ignore-platform-req=php`. What do you think?

snapshotpl:
Yeah, I forget about console deprecation...

PR #760, opened 2021-12-21
Removed integration with laminas-console and laminas-mvc-console

driehle:
**Feature Removal**

This PR removes the integration with laminas-console and laminas-mvc-console. Both packages have been deprecated and are not further maintained. The relevant signatures inside the DoctrineModule have been marked as deprecated since 4.2.0, now it's time to remove them.

Targeted at the next major release, i.e. 5.0.0.


src/Module.php
```diff
@@ -23,8 +23,6 @@
  */
 class Module implements ConfigProviderInterface, InitProviderInterface, BootstrapListenerInterface
 {
-    use GetConsoleUsage;
-
     /** @var ServiceLocatorInterface */
     private $serviceManager;
```
snapshotpl: After remove `GetConsoleUsage` can be removed

snapshotpl (changes requested):

src/Module.php
driehle: Hm, the file src/GetConsoleUsage.php is already removed, or what do you mean?

SenseException: The property doesn't seem to be in use anymore. And what about `onBootstrap()`?

driehle: Ah, I see. Fixed with 4f5dcd9.

TomHAnderson (approved):

PR #761, opened 2021-12-21
Require laminas-form 3.x, drop support for laminas-form 2.x

driehle:
This removes the polyfills introduced in #731 for the next major release.

As a result, laminas-form 3.x will be required. Support for laminas-form 2.x will be dropped.


.github/workflows/continuous-integration.yml
```diff
@@ -21,20 +21,15 @@ jobs:
           - "8.1"
         dependencies:
           - "highest"
-        laminas-form-version:
```
SenseException, line 24: The step "Upload coverage file" still uses `laminas-form-version`

SenseException (changes requested):

.github/workflows/continuous-integration.yml
driehle: Thanks, rebased and fixed in e47bb55.

TomHAnderson (approved):

PR #762, opened 2021-12-22
Require laminas-cache 3.x, drop support for laminas-cache 2.x

driehle:
This drops support for laminas-cache 2.x, so that we can get rid of `--ignore-platform-req=php` in our CI pipeline. 

With laminas-cache 3.x we have full PHP 8 support, i.e. the module can be installed without any hassle again.


TomHAnderson (approved):

driehle:
Rebased and merged

PR #765, opened 2021-12-23
Introduce typed properties where possible

driehle:
**Caution:** BC Break

This PR is a follow-up on #764  updates the coding standards to PHP 7.4 as the minimum supported version. This introduced typed properties in many places.

### DIFF
--- a/src/Authentication/Adapter/ObjectRepository.php
+++ b/src/Authentication/Adapter/ObjectRepository.php
@@ -31,7 +31,7 @@ class ObjectRepository extends AbstractAdapter
      *
      * @var mixed[]
      */
-    protected array $authenticationResultInfo = null;
+    protected ?array $authenticationResultInfo = null;
 
     protected Inflector $inflector;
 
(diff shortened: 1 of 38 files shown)


src/Authentication/Adapter/ObjectRepository.php
```diff
@@ -24,18 +24,16 @@
  */
 class ObjectRepository extends AbstractAdapter
 {
-    /** @var AuthenticationOptions */
-    protected $options;
+    protected AuthenticationOptions $options;
```
Ocramius, line 27: For `protected` stuff I suggest being extremely careful, as it can break in child classes

Ocramius (dismissed):

src/Authentication/Adapter/ObjectRepository.php
driehle: Yes, but this is targeted at 5.0.0, i.e. the next major release. Therefore, I think we should be willing to accept BC breaks in favour of code quality improvements.

Ocramius: LGTM then :+1:

TomHAnderson (approved):

SenseException (approved):

PR #766, opened 2021-12-23
Mark `Module`, `ConfigProvider`, option classes and factories as final

driehle:
**Caution:** BC Break

The `Module` and `ConfigProvider` classes, as well as the option classes and all factories are no valid extension points. They are now marked as final.

### DIFF
--- a/src/Validator/Service/AbstractValidatorFactory.php
+++ b/src/Validator/Service/AbstractValidatorFactory.php
@@ -20,10 +20,8 @@ use function sprintf;
  *
  * @internal
  */
-// phpcs:disable SlevomatCodingStandard.Classes.SuperfluousAbstractClassNaming
 abstract class AbstractValidatorFactory implements FactoryInterface
 {
-// phpcs:enable SlevomatCodingStandard.Classes.SuperfluousAbstractClassNaming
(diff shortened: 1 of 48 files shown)


TomHAnderson (approved):

src/Validator/Service/AbstractValidatorFactory.php
```diff
@@ -16,9 +16,9 @@
 use function sprintf;
 
 /**
- * Factory for creating NoObjectExists instances
+ * Base validator factory to be extended
```
SenseException: If you want a description for the class, I recommend to remove the extended part.

```suggestion
 * Base validator factory
```

1. It's abstract so it needs to be extended
2. If it's internal, it shouldn't invite to be extended outside of the DoctrineModule

SenseException (changes requested):

src/Validator/Service/AbstractValidatorFactory.php
driehle: fixed with a62b59b

driehle:
Rebased on current `5.0.x`

SenseException (approved):

PR #775, opened 2022-04-20
#774 switching to new AnnotationDriver

### LINKED ISSUE #774: incompatible with doctrine/orm >= 2.12.0

when paths are checked

ppaulis:
Signed-off-by: Pascal Paulis <ppaulis@gmail.com>

fixes #774 
linked to https://github.com/doctrine/orm/issues/9668

### DIFF
--- a/src/Service/DriverFactory.php
+++ b/src/Service/DriverFactory.php
@@ -6,7 +6,7 @@ namespace DoctrineModule\Service;
 
 use Doctrine\Common\Annotations;
 use Doctrine\ORM\Mapping\Driver\AttributeDriver;
-use Doctrine\Persistence\Mapping\Driver\AnnotationDriver;
+use Doctrine\ORM\Mapping\Driver\AnnotationDriver;
 use Doctrine\Persistence\Mapping\Driver\DefaultFileLocator;
 use Doctrine\Persistence\Mapping\Driver\FileDriver;
 use Doctrine\Persistence\Mapping\Driver\MappingDriver;
--- a/src/Service/DriverFactory.php
+++ b/src/Service/DriverFactory.php
@@ -5,8 +5,10 @@ declare(strict_types=1);
 namespace DoctrineModule\Service;
 
 use Doctrine\Common\Annotations;
+use Doctrine\ODM\MongoDB\Mapping\Driver\AnnotationDriver as AnnotationDriverODM;
+use Doctrine\ORM\Mapping\Driver\AnnotationDriver as AnnotationDriverORM;
 use Doctrine\ORM\Mapping\Driver\AttributeDriver;
-use Doctrine\ORM\Mapping\Driver\AnnotationDriver;
+use Doctrine\Persistence\Mapping\Driver\AnnotationDriver as AnnotationDriverPersistence;
 use Doctrine\Persistence\Mapping\Driver\DefaultFileLocator;
 use Doctrine\Persistence\Mapping\Driver\FileDriver;
 use Doctrine\Persistence\Mapping\Driver\MappingDriver;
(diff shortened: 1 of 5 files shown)


src/Service/DriverFactory.php
```diff
@@ -6,7 +6,7 @@
 
 use Doctrine\Common\Annotations;
 use Doctrine\ORM\Mapping\Driver\AttributeDriver;
-use Doctrine\Persistence\Mapping\Driver\AnnotationDriver;
+use Doctrine\ORM\Mapping\Driver\AnnotationDriver;
```
greg0ire: Isn't this condition supposed to work for Doctrine ODM also? Maybe the previous FQCN should still be used in the condition, and former child classes be imported as well and use in the condition as well?

greg0ire: Maybe a more robust way to do this would be to use reflection to get the number of parameters? Something like `(new ReflectionClass($class))->getConstructor()->getNumberOfParameters()`?

ppaulis: @greg0ire About your second proposition, interesting thought, but wouldn't that be a bit "too implicit"? Perhaps prone to future errors?

By using the previous FQCN, won't it be easier to maintain this place in the code, once the already deprecated driver is completely dropped?

greg0ire: See #776 for my implementation of this

I think I see what you mean: if there are other drivers with 2 arguments, it might break those. But right now, such drivers are necessarily not handled by the code, since the only other case handles only one argument.

Using the previous FQCN would help with being compatible with previous versions of the ORM, but you would have to add at least 2 lines similar to the one that uses the persistence driver class in the if condition to add support for ORM and ODM… a bit unwieldy, although more explicit. Feel free to update your PR with that.

ppaulis: > but you would have to add at least 2 lines similar to the one that uses the persistence driver class in the if condition to add support for ORM and ODM

OK, I see. I didn't think about the ODM part. Still a lot to learn about the Doctrine internals. I'll update the PR, and then you guys can decide which version you like better :+1: 

I also thought about an interface like `AnnotationDriverInterface` with a corresponding trait containing a setter for the reader. But that would mean changing the constructors. That would certainly be a more complex change.

ppaulis: @greg0ire updated the PR and added a test case for all 3 Annotation drivers (Persistence, ODM, ORM). The call to `method_exists($driver, 'getReader')` is necessary because `__invoke()` returns an instance of `MappingDriver` and not the definitive class.

As the deprecated Persistence-version is abstract and no longer parent to any specific driver, I created dummy classes for this one.

As for the `doctrine.cache.array` error you encountered in #776, this comes from the `Options\Driver` class where `$cache` is set to `array` by default. And so it tries to find a service `doctrine.cache.array` in the newly created ServiceManager.

Feel free to let me know what you think about this :+1:

src/Service/DriverFactory.php
```diff
@@ -73,7 +75,11 @@ protected function createDriver(ContainerInterface $container, Driver $options):
         // Special options for AnnotationDrivers.
         if (
             $class !== AttributeDriver::class &&
-            ($class === AnnotationDriver::class || is_subclass_of($class, AnnotationDriver::class))
+            (
+                ($class === AnnotationDriverPersistence::class || is_subclass_of($class, AnnotationDriverPersistence::class)) ||
+                ($class === AnnotationDriverORM::class || is_subclass_of($class, AnnotationDriverORM::class)) ||
+                ($class === AnnotationDriverODM::class || is_subclass_of($class, AnnotationDriverODM::class))
```
derrabus, line 81:
```suggestion
                is_a($class, AnnotationDriverPersistence::class, true) ||
                is_a($class, AnnotationDriverORM::class, true) ||
                is_a($class, AnnotationDriverODM::class, true)
```

greg0ire:
Copy pasting my comment from #776 : Now that https://github.com/doctrine/orm/pull/9671 is merged, this MR makes sense only after `doctrine/persistence` 3 is supported. Reworking the `DriverFactory` more in depth before adding support might be better than merging this.

Thank you and let's reopen this later if necessary.

PR #776, opened 2022-04-20
Adapt to new type hierarchy

### LINKED ISSUE #774: incompatible with doctrine/orm >= 2.12.0

greg0ire:
In recent versions of ORM and ODM, annotation driver no longer extend a
base class in doctrine/persistence. Instead of testing for that base
class, detecting the number of constructor arguments seems like a more
robust way to figure out what arguments to pass.

Fixes #774 

Alternative to https://github.com/doctrine/DoctrineModule/pull/775

### DIFF
--- a/tests/Service/DriverFactoryTest.php
+++ b/tests/Service/DriverFactoryTest.php
@@ -4,6 +4,9 @@ declare(strict_types=1);
 
 namespace DoctrineModuleTest\Service;
 
+use Doctrine\Common\Annotations\Reader;
+use Doctrine\Common\Cache\ArrayCache;
+use Doctrine\ORM\Mapping\Driver\AnnotationDriver;
 use Doctrine\ORM\Mapping\Driver\AttributeDriver;
 use Doctrine\Persistence\Mapping\Driver\MappingDriverChain;
 use DoctrineModule\Service\DriverFactory;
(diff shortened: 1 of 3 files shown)


greg0ire:
I tried to cover my change with a new test, but am having a hard time figuring out what I'm doing wrong. I'm getting the following error:

> Laminas\ServiceManager\Exception\ServiceNotFoundException: Unable to resolve service "doctrine.cache.array" to a factory; are you certain you provided it during configuration?

ppaulis:
> I tried to cover my change with a new test, but am having a hard time figuring out what I'm doing wrong. I'm getting the following error:
> 
> > Laminas\ServiceManager\Exception\ServiceNotFoundException: Unable to resolve service "doctrine.cache.array" to a factory; are you certain you provided it during configuration?

The following will do the trick in your test:
```
$serviceManager->setService('doctrine.cache.array', new \Doctrine\Common\Cache\ArrayCache());
```

tests/Service/DriverFactoryTest.php
```diff
@@ -90,4 +93,29 @@ public function testCreateAttributeDriver(): void
...
+        );
+        $serviceManager->setService(
+            'doctrine.cache.array',
+            new ArrayCache()
+        );
+
+        $factory = new DriverFactory('testDriver');
+        $driver  = $factory->__invoke($serviceManager, AnnotationDriver::class);
+        $this->assertInstanceOf(AnnotationDriver::class, $driver);
+        assert($driver instanceof AnnotationDriver);
```
greg0ire: I don't understand why this is necessary, but it is. PHPstan and phpstan-phpunit are installed and up to date, yet `$driver` is not narrowed down to `AnnotationDriver`.

ppaulis: I could have solved that with a simple comment:
```
/** @var AnnotationDriverORM|AnnotationDriverODM|AnnotationDriverPersistence $driver */
$driver  = $factory->__invoke($serviceManager, AnnotationDriver::class);
```

but PHPCS forced me to do the `assert(...)`. This seemed to be the only way to make both tools happy.

greg0ire: I would expect phpstan not to need this thanks to line 117. I think that works in my other projects, so there must be a configuration issue somewhere. It's mentioned in this README: https://github.com/phpstan/phpstan-phpunit

greg0ire: Found the reason: we were only using that extension to add extra rules, not to add functionality.

greg0ire:
Now that https://github.com/doctrine/orm/pull/9671 is merged, this MR makes sense only after `doctrine/persistence` 3 is supported. Reworking the `DriverFactory` more in depth before adding support might be better than merging this.

PR #777, opened 2022-04-22
Cover annotation driver instanciation

greg0ire:
It will break when allowing Persistence 3.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -12,5 +12,6 @@ parameters:
             message: '#Property DoctrineModule\\Module::\$serviceManager is never read, only written.#'
             path: src/Module.php
 includes:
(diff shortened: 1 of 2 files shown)


derrabus (approved):

driehle:
Not sure if we should push this in 4.4.x first (which is officially still supported, but will never receive support for persistence 3.0). Technically, PRs should go in the lowest supported branch, but I'll agree that this change will not be relevant before 5.2.0.

greg0ire:
It's not a big deal for me to retarget I think :)

driehle (approved):

driehle:
Looks good! Let's merge this an PR #778 to make a 4.4.2 release.

PR #778, opened 2022-04-23
Mark doctrine/orm 2.12.0 as conflict

### LINKED ISSUE #774: incompatible with doctrine/orm >= 2.12.0

driehle:
ORM 2.12.0 has a BC break which is fixed with 2.12.1. This change marks 2.12.0 as a conflict, to prevent installation.

Fixes #774

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -68,7 +68,7 @@
     },
     "require-dev": {
         "doctrine/coding-standard": "^9.0.0",
-        "doctrine/orm": "^2.10.2",
+        "doctrine/orm": "^2.10.5",
         "jangregor/phpstan-prophecy": "^1.0.0",
         "laminas/laminas-cache-storage-adapter-blackhole": "^1.2.1 || ^2.0.0",
         "laminas/laminas-cache-storage-adapter-memory": "^1.0.1 || ^2.0.0",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -85,6 +85,9 @@
         "predis/predis": "^1.1.9",
         "vimeo/psalm": "^4.11.2"
     },
+    "conflict": {
+        "doctrine/orm": "2.12.0"
+    },
```
greg0ire, line 90: Not sure this is a great idea… previous versions of the module still allow installing 2.12.0, and this might force Composer to downgrade the module to avoid this conflict, which would be bad if the module ends up getting a fix for the issue in 2.12.0. That would happen while adding compatibility with Persistence 3.

composer.json
```diff
@@ -68,7 +68,7 @@
     },
     "require-dev": {
         "doctrine/coding-standard": "^9.0.0",
-        "doctrine/orm": "^2.10.2",
+        "doctrine/orm": "^2.10.5",
```
greg0ire, line 71: What is this about?

driehle: Updating dependencies, here orm 2.10 to the latest bugfix version

composer.json
driehle: Technically, people should have a version constraint for all packages that they are using symbols of. So ORM uses should have `doctrine/orm` explicitly in their composer.json, not requiring it indirectly through DoctrineModule. They should lift this contstraint whenever the upgrade ORM, so a downgrade should not happen. However, in practice many people probably won't do so, so yeah...

Nevertheless, I think we should still add this conflict, as it is a known incompatibility. The 4.x series will not receive an update for Persistance 3, in the 5.x series support for Persistance 3 will likely come with the next minor release, i.e. 5.2.0. Since DoctrineModule 5.1.0 already requires ORM `^2.11.1`, we can probably for 5.2.0 just go with `^2.12.1` and - then - drop the constraint.

greg0ire (approved):

derrabus (approved):

PR #779, opened 2022-04-23
Remove unneeded autoload calls

greg0ire:
These were introduced in order to be able to support both persistence 1
and 2.

See #694


driehle (commented):
Good catch, thanks!

driehle (approved):
Good catch, thanks!

PR #787, opened 2022-06-21
Allow doctrine/persistence 3

snapshotpl:

PR #788, opened 2022-10-27
Add compatibility for persistence v3 with Mongo ODM

driehle:
Unfortunatly, the changes in #787 were incomplete and did not handle the instantiation of Mongo ODM drivers correctly. This leads to the error shown in doctrine/DoctrineMongoODMModule#261. The error was introduced in DoctrineModule 5.2.0. This PR fixes the mentioned error.

### DIFF
--- a/src/Service/DriverFactory.php
+++ b/src/Service/DriverFactory.php
@@ -20,7 +20,7 @@ use RuntimeException;
 
 use function class_exists;
 use function get_class;
-use function is_subclass_of;
+use function is_a;
 use function sprintf;
 
 /**
@@ -76,10 +76,8 @@ final class DriverFactory extends AbstractFactory
             $class !== ORMAttributeDriver::class &&
             $class !== MongoODMAttributeDriver::class &&
             (
-                $class === ORMAnnotationDriver::class ||
-                $class === MongoODMAnnotationDriver::class ||
-                is_subclass_of($class, ORMAnnotationDriver::class) ||
-                is_subclass_of($class, MongoODMAnnotationDriver::class)
+                is_a($class, ORMAnnotationDriver::class, true) ||
+                is_a($class, MongoODMAnnotationDriver::class, true)
             )
         ) {
             $reader = new Annotations\AnnotationReader();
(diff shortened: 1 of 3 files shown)


src/Service/DriverFactory.php
```diff
@@ -71,8 +73,14 @@ protected function createDriver(ContainerInterface $container, Driver $options):
...
-            $class !== AttributeDriver::class &&
-            ($class === AnnotationDriver::class || is_subclass_of($class, AnnotationDriver::class))
+            $class !== ORMAttributeDriver::class &&
+            $class !== MongoODMAttributeDriver::class &&
+            (
+                $class === ORMAnnotationDriver::class ||
+                $class === MongoODMAnnotationDriver::class ||
+                is_subclass_of($class, ORMAnnotationDriver::class) ||
+                is_subclass_of($class, MongoODMAnnotationDriver::class)
+            )
```
Ocramius: Perhaps this stuff can be simplified through `is_a()`? :thinking:

Ocramius (approved):

src/Service/DriverFactory.php
driehle: Good idea, I added the suggested change!

src/Service/DriverFactory.php
```diff
@@ -5,8 +5,10 @@
 namespace DoctrineModule\Service;
 
 use Doctrine\Common\Annotations;
-use Doctrine\ORM\Mapping\Driver\AnnotationDriver;
-use Doctrine\ORM\Mapping\Driver\AttributeDriver;
+use Doctrine\ODM\MongoDB\Mapping\Driver\AnnotationDriver as MongoODMAnnotationDriver;
```
malarzm, line 8: Nitpicking but MongoDB prefers to be called that as far as I know, not "Mongo" :)

malarzm: Or I think you can freely use `ODMAnnotationDriver`, we have no other actively maintained ODM anyway

malarzm (commented):
Changes LGTM but a test like https://github.com/doctrine/DoctrineModule/blob/5.3.x/tests/Service/DriverFactoryTest.php#L97 would be beneficial

PR #792, opened 2022-11-18
Drop container-interop/container-interop

driehle:
This PR drops the dependency towards `container-interop/container-interop`, which is abandonned.

As a consequence, the interface `Interop\Container\ContainerInterface` does not exist anymore in favour of `Psr\Container\ContainerInterface`. We follow the guidelines of laminas, which considers this to be a non-breaking change (see [release notes of servicemanager 3.11.0](https://github.com/laminas/laminas-servicemanager/releases/tag/3.11.0)).

All affected signatures (besides unit tests) are internal or final, so there will be no issues for users.


composer.json
```diff
@@ -37,7 +37,6 @@
     "homepage": "http://www.doctrine-project.org/",
     "require": {
         "php": "^7.4 || ~8.0.0 || ~8.1.0",
-        "container-interop/container-interop": "^1.2.0",
```
Ocramius, line 40: `psr/container` should become a dependency here, since we still reference it in `src/`

phpstan.neon
```diff
@@ -4,6 +4,10 @@ parameters:
     paths:
         - src
         - tests
+    ignoreErrors:
+        -
+          message: '#internal(G|S)etItem\(\) has parameter \$(casToken|value) with no type specified#'
+          path: src/Cache/DoctrineCacheStorage.php
```
Ocramius, line 10: :thinking: 

Is it possible for us to declare `mixed`, perhaps?

Ocramius (changes requested):

composer.json
driehle: Indeed, added

phpstan.neon
driehle: Unfortunately, no. I we add `mixed`, we loose PHP 7.4 compatibility, which I don't want to loose yet. I'd drop PHP 7.4 probably early 2023.

Ocramius: FWIW, PHP 7.4 is dead **this month**.

Anyway, you maintain this, so it's your call :+1:

Ocramius (approved):

phpstan.neon
driehle: I know, I'd consider it dead for year already, but doctrine/orm is still php: ^7.1 || ^8.0, so I'll try to be not too restrictive with DoctrineModule 🙈 

My plan is rather to make a new major release somewhere next year, then including BC breaks switching more things to PHP 8+ syntax. Probably around Q3 '23 and then going PHP 8.1+ directly...

PR #795, opened 2022-12-09
Deprecate decorating adapters from doctrine/cache for laminas-cache

driehle:
The library [doctrine/cache](https://github.com/doctrine/cache) is deprecated and no longer maintained. Therefore, uses should no longer use cache adapters from doctrine/cache, but rather adapters from [laminas-cache](https://docs.laminas.dev/laminas-cache/), [symfony/cache](https://symfony.com/doc/current/components/cache.html) or other implementations compatible to PSR-6 or PSR-16.

This PR deprecateds the class `DoctrineModule\Cache\DoctrineCacheStorage`. That class is actually a decorator, decorating a doctrine/cache adapter to be used as an laminas-cache adapter. Since usage of doctrine/cache is deprecated, usage of this decorator is now deprecated as well.

### DIFF
--- a/src/Cache/DoctrineCacheStorage.php
+++ b/src/Cache/DoctrineCacheStorage.php
@@ -10,6 +10,8 @@ use Laminas\Cache\Storage\Adapter\AdapterOptions;
 
 /**
  * Bridge class that allows usage of a Doctrine Cache Storage as a Laminas Cache Storage
+ *
+ * @deprecated 5.3.0 Usage of cache adapters from doctrine/cache is deprecated, please switch to laminas-cache.
  */
 class DoctrineCacheStorage extends AbstractAdapter
 {
--- a/composer.json
+++ b/composer.json
@@ -70,7 +70,7 @@
         "laminas/laminas-session": "^2.13.0",
         "phpstan/phpstan": "^1.9.2",
         "phpstan/phpstan-phpunit": "^1.3.0",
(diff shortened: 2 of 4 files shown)


src/Cache/DoctrineCacheStorage.php
```diff
@@ -25,6 +31,11 @@ public function __construct($options, Cache $cache)
     {
         parent::__construct($options);
 
+        trigger_error(
```
Ocramius: IMO shouldn't cause runtime side effects: this is effectively a BC break

Ocramius (changes requested):

src/Cache/DoctrineCacheStorage.php
driehle: Hm, I get your point though I am not really sure if I agree 🙈 

Anyways, let's keep it without `trigger_error()` then.

Ocramius (approved):
LGTM, thanks @driehle!

src/Cache/DoctrineCacheStorage.php
Ocramius: As someone that spen(t|ds) months in adjusting to deprecation issues in upstream, thanks for rolling this back :+1: 

`trigger_error()` is just a broken construct :)

driehle: 🤷‍♂️

PR #797, opened 2022-12-12
Psalm 5 Support

rogervila:
Hello,

This PR adds support for Psalm 5.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -72,7 +72,7 @@
         "phpstan/phpstan-phpunit": "^1.3.0",
         "phpunit/phpunit": "^9.5.27",
         "predis/predis": "^1.1.10",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -72,7 +72,7 @@
         "phpstan/phpstan-phpunit": "^1.3.0",
         "phpunit/phpunit": "^9.5.27",
         "predis/predis": "^1.1.10",
-        "vimeo/psalm": "^4.30.0"
+        "vimeo/psalm": "^4.30 || ^5.0"
```
driehle: What's the rationale behind allowing two different versions here?

rogervila: Hi, you are right, modifying it.

driehle (approved):

Issue #800, opened 2023-03-02
Can't install due to conflicting requirements


andremartinez:
I'm trying to require DoctrineModule using:
```console
composer require doctrine/doctrine-module:5.3.0
```
Pretty standard, but I'm getting the following:
```console
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires doctrine/doctrine-module 5.3.0 -> satisfiable by doctrine/doctrine-module[5.3.0].
    - doctrine/doctrine-module 5.3.0 requires doctrine/annotations ^1.13.3 -> found doctrine/annotations[1.13.3, ..., 1.14.3] but the package is fixed to 2.0.1 (lock file version) by a partial update and that version does not match. Make sure you list it as an argument for the update command.
```
The conflicting package is [PHP-CS-Fixer](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/v3.14.4/composer.json), that requires this:
```composer
        "doctrine/annotations": "^2",
```
While the [composer.json in DoctrineModule](https://github.com/doctrine/DoctrineModule/blob/5.3.0/composer.json) requires this:
```composer
        "doctrine/annotations": "^1.13.3",
```
I'm not sure whether [doctrine/annotations](https://github.com/doctrine/annotations) has a stable release for version 2, and this repo should upgrade the requirement or not

What do you guys suggest?

demiankatz:
I'm seeing the same problem and would be interested to hear a solution! I'm not familiar enough with Doctrine to know whether this module could add `|| ^2` to the annotations dependency, or if the backward-incompatible changes in annotations v2 will break something.

andremartinez:
I was thinking of forking this myself and making some tests, or even creating a pull request, but I really would like to hear someone else's opinion before committing to implementing anything

greg0ire:
Please go ahead. v2 shouldn't be too hard to upgrade to.

demiankatz:
Thanks, @andremartinez! If I can do anything to help you, please let me know. I'm not an expert in any of this (I discovered this problem because I'm in the very early stages of migrating my project from Laminas\Db to Doctrine), but still happy to help with testing, troubleshooting, etc. to the best of my ability.

andremartinez:
Ok this is gonna be one of those...

The `require-dev` dependency of `doctrine/mongodb-odm` also requires `"doctrine/annotations": "^1.*",` on their `require` meaning an upgrade should happen there as well.

[...]

PR #801, opened 2023-03-06
Adjust dependencies/code for doctrine/annotations 2 compatibility

demiankatz:
This PR is work in progress on addressing #800.

TODO
- [x] Resolve cache-related issues
- [x] Test with doctrine/annotations 1.x after finishing code changes to see if backward-compatibility is maintained, or if we need to fully bump the required version

### DIFF
--- a/src/Module.php
+++ b/src/Module.php
@@ -18,11 +18,6 @@ final class Module implements ConfigProviderInterface, InitProviderInterface
 {
     public function init(ModuleManagerInterface $manager): void
     {
-        AnnotationRegistry::registerLoader(
-            static function ($className) {
-                return class_exists($className);
-            }
-        );
     }
 
     /**
--- a/src/Module.php
+++ b/src/Module.php
@@ -5,18 +5,12 @@ declare(strict_types=1);
 namespace DoctrineModule;
 
 use Laminas\ModuleManager\Feature\ConfigProviderInterface;
-use Laminas\ModuleManager\Feature\InitProviderInterface;
-use Laminas\ModuleManager\ModuleManagerInterface;
 
 /**
  * Base module for integration of Doctrine projects with Laminas applications
  */
-final class Module implements ConfigProviderInterface, InitProviderInterface
+final class Module implements ConfigProviderInterface
 {
-    public function init(ModuleManagerInterface $manager): void
-    {
-    }
-
     /**
      * @return array<string, mixed>
      */
--- a/src/Module.php
+++ b/src/Module.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace DoctrineModule;
 
-use Doctrine\Common\Annotations\AnnotationRegistry;
 use Laminas\ModuleManager\Feature\ConfigProviderInterface;
 use Laminas\ModuleManager\Feature\InitProviderInterface;
 use Laminas\ModuleManager\ModuleManagerInterface;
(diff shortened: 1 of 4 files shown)


greg0ire:
Regarding the cache, there is already an issue about this it seems: https://github.com/doctrine/DoctrineModule/issues/786

demiankatz:
Thanks, @greg0ire -- in fact, it looks like there's work in progress at #796.

demiankatz:
I did some experimentation with merging together this PR and #786, but it still doesn't work -- I've left some more detailed comments on #786 since I think it might make sense to try to get the cache updates done before tackling this part.

demiankatz:
@greg0ire, I found a little more time to work with this further today, and I've gotten this into a state that passes all tests with both annotations v1 and annotations v2. However, when using annotations v2, caching functionality is going to be lost (for now) because the cache objects do not comply with PSR-6. However, I believe this is written in such a way that if we return PSR-6 compliant objects in future, the caching will begin to work -- I'm not sure if #796 can help with that.

Obviously, as noted earlier, I don't have a deep understanding of this module, so it's possible I'm overlooking something important... but hopefully passing tests are a good first step. Please let me know if there's more I can do to help! I've taken this PR out of draft mode so that it can be reviewed.

src/Module.php
```diff
@@ -4,25 +4,17 @@
...
  */
 final class Module implements ConfigProviderInterface, InitProviderInterface
 {
     public function init(ModuleManagerInterface $manager): void
     {
-        AnnotationRegistry::registerLoader(
-            static function ($className) {
-                return class_exists($className);
-            }
-        );
```
greg0ire, line 25: Not a Zend specialist, but should `init()` be removed entirely?

demiankatz: You're right, this method is not required, so I have just removed it (and several now-obsolete interfaces).

demiankatz:
In reviewing #800, I was reminded that @andremartinez was encountering phpstan failures. I was able to reproduce the problem and found that it was caused by referencing a non-existent (legacy) Blackhole cache adapter in a test. The easiest solution was to simply skip the test if the newer, expected version of the adapter was missing. I'm not seeing any skipped tests when I run the suite in my environment, so I think this is probably safe for now, especially since things are going to be significantly revised by #796 when that work is completed.

demiankatz:
Apologies, I see that my phpstan fix broke the styles. Everything should be corrected now!

greg0ire (approved):

demiankatz:
Thanks, @greg0ire!

andremartinez (approved):
There are warnings for some lines in the Driver Factory, but those doesn't look like breaking changes. Proper unit testing should be added.

Other than that, everything looks good to me

demiankatz:
Thanks, @andremartinez! It may make sense to finish the cache related work before addressing the test coverage warnings. Some lines may not be reachable until that's done, and others may become obsolete. Let me know if I can do more to help, in any case!

TomHAnderson:
The order of PRs Dennis and I are shooting for is 

#799 
#801 
#796 

@driehle Are you back and ready for #799?

demiankatz:
Thanks, @TomHAnderson, let me know if I can do anything else to help (either in terms of contributing to #799, or in terms of improving test coverage in a new PR after #796 is merged). My time is going to be pretty limited for the next couple of weeks, but if I plan ahead, I can help out a little down the road. :-)

TomHAnderson:
In the interest of moving this forward I've deprioritized #799 at this time though I expect to merge it before a release is made.

PR #804, opened 2023-04-18
Feat/merge drop php74

TomHAnderson:
This is a continuation of #799 with fixes for lastest code

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -1,6 +1,8 @@
 <?xml version="1.0"?>
 <psalm
     errorLevel="4"
+    findUnusedCode="false"
+    findUnusedBaselineEntry="true"
     resolveFromConfigFile="true"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xmlns="https://getpsalm.org/schema/config"
--- a/src/Authentication/Adapter/ObjectRepository.php
+++ b/src/Authentication/Adapter/ObjectRepository.php
@@ -12,7 +12,6 @@ use Laminas\Authentication\Adapter\Exception;
 use Laminas\Authentication\Result as AuthenticationResult;
 
 use function call_user_func;
-use function get_class;
 use function method_exists;
 use function property_exists;
 use function sprintf;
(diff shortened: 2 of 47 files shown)


psalm.xml
```diff
@@ -1,6 +1,8 @@
 <?xml version="1.0"?>
 <psalm
     errorLevel="4"
+    findUnusedCode="false"
```
Ocramius, line 4: Should probably be `true` and expand the baseline

src/Authentication/Adapter/ObjectRepository.php
```diff
@@ -38,7 +37,7 @@ class ObjectRepository extends AbstractAdapter
      *
      * @param mixed[]|AuthenticationOptions $options
      */
-    public function __construct($options = [], ?Inflector $inflector = null)
+    public function __construct(array|AuthenticationOptions $options = [], ?Inflector $inflector = null)
```
Ocramius, line 40: I'm OK with adding type declarations, but be very careful with added native types, since you may run into BC breaks

src/Validator/ObjectExists.php
```diff
@@ -43,7 +43,7 @@ class ObjectExists extends AbstractValidator
      *
      * @var mixed[]|string
      */
-    protected $fields = null;
+    protected mixed $fields = null;
```
Ocramius, line 44: This type change seems invalid

Ocramius (changes requested):

src/Authentication/Adapter/ObjectRepository.php
TomHAnderson: A new major version, 6.0.x, allows for BC breaks

psalm.xml
TomHAnderson:
```
When true, Psalm will attempt to find all unused code (including unused variables), the equivalent of running with --find-unused-code. Defaults to false.
```

This value will default to true in psalm 6.  I want to leave it as-is at this time because there is a lot of unused code in the library (a result of incomplete unit tests?)

src/Validator/ObjectExists.php
TomHAnderson: The fields property is typecast to an array in `validateFields()` thereby allowing objects with `ArrayAccess`. 

This object is not a final class so keeping this as `mixed` is appropriate for typed properties.

TomHAnderson: Annotation dropped in d3ea2f98bbaed8bdbf666674543f0819e77bdeed

PR #806, opened 2023-06-08
Fix cache configuration for better Doctrine compatibility.

demiankatz:
The default Laminas cache key pattern is incompatible with Doctrine's basic assumptions. This PR loosens things up a bit. It might not be perfect -- I'm open to suggestions for cleaner regexes -- but it at least works with the current DoctrineORMModule test suite.

Additionally, the old Doctrine file system cache adapter included automatic serialization, but the Laminas file system cache adapter does not. This adjusts the configuration to fix that inconsistency and ensure compatible behavior.

See https://github.com/doctrine/DoctrineORMModule/pull/734 for related discussion.

### DIFF
--- a/src/ConfigProvider.php
+++ b/src/ConfigProvider.php
@@ -117,33 +117,34 @@ final class ConfigProvider
      */
     public function getCachesConfig(): array
     {
+        $defaultOptions = [
+            'namespace' => 'DoctrineModule',
+            'key_pattern' => '/^[a-z0-9_\+\-\[\]\\\\$]*$/Di'
+        ];
         return [
             'doctrinemodule.cache.apcu' => [
                 'adapter' => 'apcu',
-                'options' => ['namespace' => 'DoctrineModule'],
+                'options' => $defaultOptions,
             ],
             'doctrinemodule.cache.array' => [
                 'adapter' => Memory::class,
-                'options' => ['namespace' => 'DoctrineModule'],
+                'options' => $defaultOptions,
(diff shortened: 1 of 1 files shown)


demiankatz:
I see there are some style issues that need addressing; I have to run for a few minutes, but I'll fix them as soon as I get back!

demiankatz:
Okay, style issues fixed -- this should be ready for review, if @greg0ire or @TomHAnderson has a moment.

TomHAnderson (approved):

demiankatz:
Thanks, @TomHAnderson! I don't expect to need to do any more work here in the immediate future, so if you don't mind releasing 6.0.2, that would be fantastic. :-)

Novynn:
This change is breaking things. An example is: annotation methods are cached by Doctrine using a hash symbol (ie. #) which isn't included in the regex.

demiankatz:
Feel free to open a new PR to further open up the regex. I'm not familiar enough with doctrine to know all the possibilities, but it was even more broken before my changes. I'd appreciate any help you can offer to fix it the rest of the way!

TomHAnderson:
Reconfiguring the caching is a big change, hence the major release.  If you can improve the regex or give examples of the problem you're seeing it would be most helpful.

demiankatz:
It would probably be a good idea to have a unit test that exercises common cache keys to prevent regressions like this. I'll see if I can help with that in the near future!

PR #809, opened 2023-06-09
Add support for # in cache key; add test.

### LINKED ISSUE #807: 6.0.2 regex change is too restrictive
This change (#806) is breaking things. An example is: annotation methods are cached by Doctrine using a hash symbol (ie. #) which isn't included in the regex.

The initial regex that was breaking the tests in https://github.com/doctrine/DoctrineORMModule/pull/734 is only relevant to the "Filesystem" storage adapter.

https://github.com/laminas/laminas-cache-storage-adapter-filesystem/blob/d78283f47a7d0fd3968c63836ee2805b53bff06f/src/FilesystemOptions.php#L27

demiankatz:
Follow-up to #806. Resolves #807.

### DIFF
--- a/src/ConfigProvider.php
+++ b/src/ConfigProvider.php
@@ -117,10 +117,7 @@ final class ConfigProvider
      */
     public function getCachesConfig(): array
     {
-        $defaultOptions = [
-            'namespace' => 'DoctrineModule',
-            'key_pattern' => '/^[a-z0-9_\+\-\[\]\\\\$]*$/Di',
-        ];
+        $defaultOptions = ['namespace' => 'DoctrineModule'];
 
         return [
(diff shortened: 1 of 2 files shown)


demiankatz:
This should be ready for merge, @TomHAnderson, unless @Novynn is aware of any additional characters that should be added to the regex and corresponding test.

TomHAnderson (approved):
I'll merge this because it's certainly broken.  More testing of individual adapters is in order.

demiankatz:
If further followup is needed, I'm happy to help, time permitting. I hope to spend some time in the next couple of weeks integrating this into my own application, so hopefully that will help turn up any remaining issues!

PR #812, opened 2023-07-17
Allow doctrine/collections:^2.1

fezfez:


SenseException (approved):
Build works with both version constraints.

PR #813, opened 2023-07-17
Allow doctrine/event:^2.0

fezfez:


SenseException (approved):
Build works with both version constraints.

I'm not sure which version this should be released with. If people update to the next 6.0 release, the change of `$this->getListeners()` will break their project and they have to set the constraint themselves (if they find out that this comes from the change in composer).

PR #821, opened 2023-10-17
Remove `provide` section from `composer.json`

### LINKED ISSUE #820: Package incorrectly states that it does provide `laminas/laminas-cache-storage-implementation`
As per `composer.json`, this package is stating that it provides `laminas/laminas-cache-storage-implementation` via `provide`.
Only packages which do actually `provide` an implementation should use `provide`.

There is a good read about dependency inversion [here](https://matthiasnoback.nl/2014/10/composer-provide-and-dependency-inversion/).

Please remove this from the projects `composer.json` to avoid future problems with incompatibilities between `laminas-cache` and `doctrine/doctrine-module`.

boesing:
This project does not provide any cache implementation and therefore should not expose this `provide` section.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -79,9 +79,6 @@
     "conflict": {
         "doctrine/orm": "2.12.0"
     },
-    "provide": {
-        "laminas/laminas-cache-storage-implementation": "1.0.0"
-    },
     "suggest": {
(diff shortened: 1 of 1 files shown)


driehle:
The Psalm error seems to be the same as in doctrine/DoctrineORMModule#745, we'll care about that separately.

driehle (approved):

PR #823, opened 2023-10-21
Fix unwanted BC break from 5.x series not allowing integer input for validators

### LINKED ISSUE #818: Allow cleanSearchValue method to accept an integer
Just to understand, why doesn't the cleanSearchValue method in the ObjectExists validator accept an integer? If the purpose of the method is to verify an entity with a given id.

Line: https://github.com/doctrine/DoctrineModule/blob/6.1.x/src/Validator/ObjectExists.php#L128

driehle:
Closes #818 

The release of the 6.x series introduced an unwanted behaviour not allowing integer input to the form validators. This needs to be changed, as integers (e.g. IDs) should be an acceptable input.

### DIFF
--- a/src/Validator/ObjectExists.php
+++ b/src/Validator/ObjectExists.php
@@ -118,14 +118,14 @@ class ObjectExists extends AbstractValidator
     }
 
     /**
-     * @param string|object|mixed[] $value a field value or an array of field values if more fields have been configured to be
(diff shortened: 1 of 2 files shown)


driehle:
We don't need to care about the Psalm error `MissingTemplateParam`, this is done in #822.

TomHAnderson (approved):

PR #828, opened 2024-01-09
Support for PHP 8.3

TomHAnderson:


src/Form/Element/ObjectMultiCheckbox.php
```diff
@@ -54,7 +54,7 @@ public function setValue($value)
     /**
      * {@inheritDoc}
      *
-     * @return array<array-key,mixed>
+     * @return array<int<0, max>|string, array{attributes?: array<string, scalar|null>, disabled?: bool, label: non-empty-string, label_attributes?: array<string, scalar|null>, selected?: bool, value: non-empty-string}|string>
```
SenseException: Maybe this can be put into a type and used with import-type. Does Psalm and PHPStan share a common understanding with an annotation for importing a type?

SenseException (approved):

driehle (changes requested):
Please update PHP version of the other workflows to 8.2, i.e. coding standards, static analysis and composer lint. They need to run with the latest version, as the minimal version is automatically inflected.

TomHAnderson:
Return types are now imported from Laminas types.  Workflows updated to 8.2

driehle (approved):
LGTM

driehle:
This must be a new minor according to our rules, i.e. 6.1.0

PR #829, opened 2024-01-19
psr/container 2.0

TomHAnderson:


SenseException:
None of the Continuous Integration builds is using psr/container 2.

driehle (changes requested):
It doesn't make any sense to support psr/container 2 unless we upgrade laminas-servicemanager to 4.0, as 3.x supports psr/container 1 only. For that, however, we will need to go into a new major release. I have changed the milestone accordingly.

Let's wait for laminas-servicemanager 4.0 to be published. An RC is available already.

TomHAnderson:
The issue for Service Manager 4 is https://github.com/laminas/laminas-cache/pull/297

TomHAnderson:
Work on updating this library will be tracked here: https://github.com/doctrine/DoctrineModule/milestone/35

PR #836, opened 2024-08-04
Feature/composer versions

TomHAnderson:
This PR is not a complete `composer.json` but it does compose.  Laminas libraries not included do not support Service Manager 4.


TomHAnderson:
Psalm has been removed because it doesn't support `nikic/php-parser` ^5.0

TomHAnderson:
Cache adapters and validators are not included in this PR.  Those are not updated to use SM4.

TomHAnderson:
Merging to 7.0.0 milestone knowing tests do not pass.

PR #862, opened 2026-03-28
Add PHP 8.5 support

leuchtdiode:


leuchtdiode:
Looks like we got bigger problems here. Just bumping the cache adapters won't work as they need laminas-cache 4.x.x which needs service manager 4.x.x. But laminas-form needs service manager 3.x.x

